PR #2, opened 2017-02-20
Ws date picker preact

fragsalat:
image


faboweb:
redid this in #3 as it should be merged into release

PR #3, opened 2017-02-21
Ws date picker preact

faboweb:
see #2 for an image of the new date picker


components/ws-date-picker/ws-date-picker-calendar-row.js
```diff
@@ -0,0 +1,43 @@
...
+		return momentDay.month() === expectedMonth;
+	}
+	isActive(year, month, date) {
+		return this.props.momentDay
+			&& year === this.props.momentDay.year()
+			&& month === this.props.momentDay.month()
+			&& date === this.props.momentDay.date();
+	}
+	render() {
+		let weeknum = this.props.weeknum;
```
fokusferit: is it maybe "weekNumber" ? to be more consistent here

fokusferit: And would use `let { weeknum, month , year} = this.props;`

package.json
```diff
@@ -53,10 +53,12 @@
     "lite-server": "^2.2.0",
     "mocha": "3.2.0",
     "mocha-webpack": "0.7.0",
+    "moment": "2.0.0",
```
fokusferit: Just curious are we using here only moment or also moment.js with locales ?  As with locales it is quite 'big' (70kb~). But I think this is something we should consider for  #7 when building it.

fragsalat: I don't like 3rdParty dependencies but if this is working I'll accept this now and mark it as to be refactored.
Funny side fact that also moment introduces new complexity in the back^^

webpack.config.js
```diff
@@ -7,24 +7,28 @@ module.exports = {
...
-        loader: 'sass-loader', // compiles Sass to CSS
-      }],
-    }],
+		rules: [{
+        test: /\.js$/,
+        use: [{
+          loader: 'babel-loader',
+          options: { presets: ['es2015'] },
+        }],
+				exclude: [/node_modules/]
```
fragsalat: Formatting sux but it's ok for now.

PR #4, opened 2017-02-23
Adding linting rules to enforce jsdocs

fokusferit:
My comments on the ruleset:

- I think no one uses Arrow Functions for creating components, classes etc. anyway so documenting them is optional. But Function declarations (Es6 class methods), Class Declarations and Methods should be documented and therefore required.

I've also added 'valid-jsdoc' check to make sure that our jsdocs are also valid. Currently, it is only defined that @return is not necessary when there is no return in the method and everything has to have a description. 
If you're interested we can also add here a 'prefer' option which forces everyone to use the same jsdoc tags http://eslint.org/docs/rules/valid-jsdoc#options

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -70,6 +70,18 @@ module.exports = {
     "react/self-closing-comp": 0,
     "react/prop-types": 0,
     "react/prefer-stateless-function": 1,
-    "require-yield": 0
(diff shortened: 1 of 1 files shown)


faboweb:
@fokusferit should be a PR to release branch

PR #6, opened 2017-02-23
WS-Notification

tonysaad:
Adding ws-notification component using preact
Usage Example:
I've changed it , it's not the same like #11 anymore as I started using Custom Events attached `window`
I just did that to make it easy to use in any different application without caring much about the `init()` or Global Variables.

1. To Create a new Notification from anywhere in the app just do the following: 
```
dispatchEvent(new CustomEvent('ws-notification-open', {
'detail': {
    title: 'string', 
    description: 'string, optional', 
    type: 'string, default: info', // other values: 'success','info','error'
    lifetime: 0 // Lifetime 0 means endless, default is 5000ms
}
}))
```

2. To close all notifications, just do the following: 
```
dispatchEvent(new CustomEvent('ws-notification-close-all'))
```


faboweb:
see https://github.bus.zalan.do/WholesaleDesignSystem/Styleguide-Web-Components/pull/49

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+    if (!type) {
+      type = DEFAULT_NOTIFICATION_TYPE;
+    }
+    this.setState({
+      notifications: this.state.notifications.concat([{title, description, type, lifetime}])
+    });
+  }
+  animateIn(notification, i) {
+    const that = this;
```
fragsalat: Again I don't understand why you specify that = this if you never use 'that'

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+    this.refs.list.style.transition = `none`;
+    this.refs.list.style.transform = `translate3d(0, 80px, 0)`;
+    setTimeout(() => {
+        this.refs.list.style.transition = `transform .35s cubic-bezier(.35, 1, .69, .98) .1s`;
+        this.refs.list.style.transform = `translate3d(0, 0, 0)`;
+    }, 0);
+    clearTimeout(this.state.timeoutId);
+    this.setState({timeoutId: setTimeout(function () {
```
fragsalat: The timeoutId shouldn't trigger a re-rendering or?

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+  componentWillUnmount() {
+    window.removeEventListener('ws-notification-open');
+    window.removeEventListener('ws-notification-close');
+  }
+  addNotify (event) {
+    let {title, description, type, lifetime} = event.detail;
+    if (typeof lifetime === 'undefined') {
+      lifetime = DEFAULT_NOTIFICATION_LIFETIME;
+    } else if (!lifetime) {
+      lifetime = 2147483647;
```
fragsalat: I liked the way before more since error notifications don't need any timeout which costs resources.

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+    setTimeout(() => {
+        this.refs.list.style.transition = `transform .35s cubic-bezier(.35, 1, .69, .98) .1s`;
+        this.refs.list.style.transform = `translate3d(0, 0, 0)`;
+    }, 0);
+    clearTimeout(this.state.timeoutId);
+    this.setState({timeoutId: setTimeout(function () {
+      return that.close(i);
+    }, notification.lifetime)});
+  }
+  closeAllEvent() {
```
fragsalat: Events not Event^^

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+    let i;
+    let len = this.state.notifications.length;
```
fragsalat: No need for a new variable here. It's just a pointer to a pointer to a value which is redundant.

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+      this.close(i);
+    }
+  }
+  close(i) {
+    if (this.refs[`notification-${i}`]) {
+      this.refs[`notification-${i}`].style.transition = 'opacity .2s ease-out, max-height .8s ease, margin-bottom .8s ease';
+      this.refs[`notification-${i}`].style.willChange = 'opacity, max-height, margin-bottom';
+      this.refs[`notification-${i}`].style.opacity = 0;
+      this.refs[`notification-${i}`].style.maxHeight = 0;
+      this.refs[`notification-${i}`].style.marginBottom = '-4.5rem';
```
fragsalat: I prefer to create a variable for this.
```js
const notification = this.refs[`notification-${i}`];
if (notification) {
  notification.style.transition = '...';
  //  ...
}
```

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+            timeoutId: null,
+            notifications
+          });
+      }, 1000);
+    }
+  }
+  render() {
+    const that = this;
+    let notificationsList = [];
+    this.state.notifications.forEach((notification, i) => {
```
fragsalat: This could be done inside the jsx using map

components/ws-notification/ws-notification.scss
```diff
@@ -0,0 +1,116 @@
...
...NhsAlDKyMgPCAuIzixKwEUKy2wQSyxCAQlQrAAFrAEJbAEJSAuRyNHI2EgsAQjQrAJQysgsGBQWCCwQFFYswIgAyAbswImAxpZQkIjIEewBEOwAmIgsABQWLBAYFlmsAFjYCCwASsgiophILACQ2BkI7ADQ2FkUFiwAkNhG7ADQ2BZsAMlsAJiILAAUFiwQGBZZrABY2GwAiVGYTgjIDwjOBshICBGI0ewASsjYTghWbErARQrLbBCLLA1Ky6xKwEUKy2wQyywNishIyAgPLAEI0IjOLErARQrsARDLrArKy2wRCywABUgR7AAI0KyAAEBFRQTLrAxKi2wRSywABUgR7AAI0KyAAEBFRQTLrAxKi2wRiyxAAEUE7AyKi2wRyywNCotsEgssAAWRSMgLiBGiiNhOLErARQrLbBJLLAII0KwSCstsEossgAAQSstsEsssgABQSstsEwssgEAQSstsE0ssgEBQSstsE4ssgAAQistsE8ssgABQistsFAssgEAQistsFEssgEBQistsFIssgAAPistsFMssgABPistsFQssgEAPistsFUssgEBPistsFYssgAAQCstsFcssgABQCstsFgssgEAQCstsFkssgEBQCstsFossgAAQystsFsssgABQystsFwssgEAQystsF0ssgEBQystsF4ssgAAPystsF8ssgABPystsGAssgEAPystsGEssgEBPystsGIssDcrLrErARQrLbBjLLA3K7A7Ky2wZCywNyuwPCstsGUssAAWsDcrsD0rLbBmLLA4Ky6xKwEUKy2wZyywOCuwOystsGgssDgrsDwrLbBpLLA4K7A9Ky2waiywOSsusSsBFCstsGsssDkrsDsrLbBsLLA5K7A8Ky2wbSywOSuwPSstsG4ssDorLrErARQrLbBvLLA6K7A7Ky2wcCywOiuwPCstsHEssDorsD0rLbByLLMJBAIDRVghGyMhWUIrsAhlsAMkUHiwARUwLQBLuADIUlixAQGOWbABuQgACABjcLEABUKyAAEAKrEABUKzCgIBCCqxAAVCsw4AAQgqsQAGQroCwAABAAkqsQAHQroAQAABAAkqsQMARLEkAYhRWLBAiFixA2REsSYBiFFYugiAAAEEQIhjVFixAwBEWVlZWbMMAgEMKrgB/4WwBI2xAgBEAAA=') format('truetype');
```
fragsalat: I know you just copied it but I still want to mention that this can be removed and the global available icons can be used in the future. Nothing to do right now

components/ws-notification/index.js
```diff
@@ -0,0 +1,110 @@
...
+      notifications: [],
+      timeoutId: null
+    };
+    // Bind functions to this scope
+    this.addNotify = this.addNotify.bind(this);
+    this.closeAllEvent = this.closeAllEvent.bind(this);
+  }
+  componentDidMount() {
+    window.addEventListener('ws-notification-open', this.addNotify);
+    window.addEventListener('ws-notification-close-all', this.closeAllEvent);
```
fragsalat: I really like this approach since the application doesn't have to care if the notification component works or not.

fragsalat:
I recognized that the target branch is wrong^^
It should be release and not preact^^

tonysaad:
@fragsalat Yeah That's totally right I will work on your comments first, and will create a new PR (I already merged release into it)

components/ws-notification/index.js
tonysaad: How would it be without any timeout ? Only on click and the `close-all` event ?

fragsalat: Don't call setTimeout and call the function directly^^

tonysaad: But this way it will always be visible until the user clicks, we'd need a timeout to hide it after some period of time.

fragsalat:
I'm fine with it. The jsdoc is missing but I'll add it later.

I'll close this PR and reopen it to the release branch.

PR #7, opened 2017-02-27
[Preliminary] Build es5 components

faboweb:
intention to build components as es5 imports so we don't rely on the build system of the individual projects these components are used in

proxying react and react-dom to preact will always be needed.

status: this is in dev phase. not stable yet.


fragsalat:
@faboweb can this be closed now?

PR #8, opened 2017-02-27
Preact dropdown

fragsalat:
I migrated the dropdown to preact and implemented the filter possibility.
Beside that I did some fixes and restructuring to fix the installation with jspm.


package.json
```diff
@@ -27,44 +27,49 @@
...
     "webpack": "2.2.1",
     "webpack-dev-server": "2.3.0",
-    "wholesale-styleguide-scss": "git+ssh://git@github.bus.zalan.do/WholesaleDesignSystem/Styleguide-SCSS.git"
+    "wholesale-styleguide-scss": "git+ssh://git@github.bus.zalan.do/WholesaleDesignSystem/Styleguide-SCSS.git#dropdown-adjustments"
   },
-  "dependencies": {
+  "peerDependencies": {
     "preact": "7.1.0",
-    "preact-compat": "3.13.1",
-    "react": "15.4.2",
```
faboweb: you removed react. but it should be there to also test for react

src/imports.js
```diff
@@ -4,6 +4,7 @@ import * as ReactDOM from 'react-dom';
 // API shared by Preact and React
 export const React = {
   createElement: _React_.createElement || _React_.h,
+  PropTypes: _React_.PropTypes || { oneOf: () => {} },
```
faboweb: oneOf is some kind of json API? could you maybe add a comment?

src/ws-dropdown-item.js
```diff
@@ -0,0 +1,126 @@
...
+   * or change the dropdown value for simple dropdown's
+   * @param event {MouseEvent}
+   * @void
+   */
+  onClick(event) {
+    event.stopPropagation();
+    // Click on parent means back navigation
+    if (this.props.isParent) {
+      // dropdown-item(go-back) -> dropdown-menu(show-parent) -> dropdown-item(show-parent) -> dropdown-menu
+      this.props.handle('go-back');
```
faboweb: would it maybe make more sense to add explicit callbacks?

webpack.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
       test: /\.js$/,
       use: [{
         loader: 'babel-loader',
-        options: { presets: ['es2015'] },
+        options: { babelrc: true },
```
faboweb: 👍

src/imports.js
faboweb: figured it is the one function you use inside of the code. a comment would be nice.

package.json
fragsalat: But then they are needed only as devDependencies but not as dependencies or peerDependencies. I gonna add them

src/imports.js
fragsalat: I'll add one. Though it's clear that this is the filler if preact is used instead of react

src/ws-dropdown-item.js
fragsalat: Though about that too but decided against it since this handle function covers up to 6 cases I don't want to add 6 callbacks to all dropdown items/menus. 
https://github.com/wholesale-design-system/components/pull/8/files/efef757429e3a37f0465dac9a46714a66c02ad47#diff-2a426cf03d9e6062ec7e0e3c2862ceeaR133

Also I wanted a consistent behaviour through all dropdown components.

src/ws-dropdown.js
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * Opens the dropdown
+   * @void
+   */
+  open() {
+    if (this.state.opened) {
+      return;
+    }
+    this.state.opened = true;
```
tonysaad: I'd use `this.setState({opened: true})` here

src/ws-dropdown.js
```diff
@@ -0,0 +1,252 @@
...
+
+  /**
+   * Closes the dropdown and clears the selection if needed
+   * @void
+   */
+  close() {
+    if (!this.state.opened) {
+      return;
+    }
+    this.state.opened = false;
```
tonysaad: I'd use `this.setState({opened: false})` here

src/ws-dropdown.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  /**
+   * Set the value of the dropdown and update the display text if the trigger element is a select
+   * @param value {Object|Array<Object>}
+   * @void
+   */
+  setValue(value) {
+    this.state.value = value;
```
tonysaad: I'd use `this.setState({value});` here

src/ws-dropdown.js
```diff
@@ -0,0 +1,252 @@
...
+    // Check if we have to update the text value
+    if (this.props.type === 'select') {
+      if (Array.isArray(this.state.value)) {
+        this.state.text = this.state.value.map(item => item.label).join(', ');
```
tonysaad: I'd assign the text value to variable and then setState ex. `this.setState({text});`

src/ws-dropdown-menu.js
```diff
@@ -0,0 +1,289 @@
...
+    this.props.handle('change', this.state.value);
+    this.setState(this.state);
+  }
+
+  /**
+   * Handles data propagation from child menus
+   * @param type {String} Should be just show-parent
+   * @param data {*}
+   */
+  handlePropagation = (type, data) => {
```
tonysaad: I love this approach, it's new to me and I totally liked it

src/ws-dropdown-item.js
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Important to update the state when the items in menu changed.
+   * When in the multi select a dropdown item is selected and stored (submitted) it will be filtered out
+   * of the item list and shown in a separate value list. React doesn't created new components but reuses them.
+   * This leads to property updates and without this function a wrong item will be rendered.
+   * @param props {Object}
+   * @void
+   */
+  componentWillReceiveProps(props) {
+    this.state = props.item;
```
tonysaad: I've checked the comment but I'm still not sure here why do we need the state, as we always have the label & children in the props and they're not changeable I'd go for using props and don't create a state if possible.

As was mentioned in React Guides:
> Using props to generate state in getInitialState often leads to duplication of “source of truth”, i.e. where the real data is. This is because getInitialState is only invoked when the component is first created.

src/ws-dropdown-item.js
```diff
@@ -0,0 +1,126 @@
...
+    let anchorClass = 'text';
+    anchorClass += this.state.selected ? ' is-active' : '';
+    anchorClass += this.state.focused ? ' is-focused' : '';
+    anchorClass += this.state.disabled ? ' is-disabled' : '';
+
+    return (
+      <li className="dropdown-item" onClick={event => this.onClick(event)}>
+        <a className={anchorClass} href={this.state.href}>
+          {this.state.icon || this.props.icon &&
+            <i className={`icon ${this.state.icon || this.props.icon}`}/>
```
tonysaad: Is this because the icon might change for the same component as you mentioned in the comments ?

src/ws-dropdown-menu.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+  };
+
+  /**
+   * Shows the child menu and hides the current menu
+   * @param subMenu {WSDropdownMenu}
+   * @void
+   */
+  showChild(subMenu) {
+    this.state.openSubMenu = subMenu;
```
tonysaad: Here I think it'd work if we use nested refs ex. `this.refs.WSDropdownItem.refs.openMenu` I'm not sure if we'd need state for this.
If there's no way around it, I'd still suggest using this.setState() and not mutate the value directly

src/ws-dropdown.js
fragsalat: True. Changes will come

src/ws-dropdown.js
fragsalat: True. Changes will come

src/ws-dropdown.js
fragsalat: Since this should not trigger re rendering I moved opened out of the state and to instance scope. Changes will come

src/ws-dropdown.js
```diff
@@ -0,0 +1,252 @@
...
+  /**
+   * Opens the dropdown
+   * @void
+   */
+  open() {
+    if (this.state.opened) {
+      return;
+    }
+    this.state.opened = true;
+    this.dropdownContainer.style.height = 0;
```
fragsalat: I have `ref={element => { this.dropdownContainer = element; }}` which should do the same. But nice to know that the other way works as well.
Can you tell me how this could happen or how to reproduce?

fragsalat: BTW react's eslint suggests to not use string references https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md

src/ws-dropdown.js
fragsalat: As mentioned above I moved this property out of the state since a re-rendering is not the goal

src/ws-dropdown-menu.js
fragsalat: I tried to iterate over all functions and to bind them to the scope but it didn't worked out because this function was on the parent class and therefore not included in for in. So i choosed this way instead of multiple .bind(this)

src/ws-dropdown-menu.js
fragsalat: I'll move this out of the state. WSDropdownItem is inside an array so accessing the ref of the item doesn't make that much sense against passing the needed ref.

src/ws-dropdown-item.js
fragsalat: I'd like to use the state since selected and stored can change and I don't want to build a functionality to propagate it. Using the item as the state the parent knows automatically whats going on because it's just a reference.

src/ws-dropdown-item.js
fragsalat: Ah I have to change the order. For parents there will be a arrow to the left displayed instead of the icon. So I call it like
```js
<WSDropdownItem
            item={this.props.parent}
            icon="icon-left"
            handle={this.handlePropagation}
            key="parent"
            isParent
          />,
```

fragsalat: Fixed

src/ws-dropdown-menu.js
fragsalat: Fixed

src/ws-dropdown.js
fragsalat: Fixed

src/ws-dropdown.js
fragsalat: Fixed the null error.

src/ws-dropdown.js
fragsalat: Fixed

src/ws-dropdown.js
fragsalat: Fixed

src/ws-dropdown.js
fragsalat: Fixed

src/imports.js
fragsalat: Fixed

package.json
fragsalat: Fixed

PR #13, opened 2017-03-10
Adding gulp tasks for generating jsdocs -> markdown API documentation.

fokusferit:
Adding gulp tasks for creating jsdocs -> markdown API documentation.
Basically:
- I'm building temporarily all components to ES2015 + comments
- Using JsdocsTool to generate Markdown out of jsdocs
- Saving them in dist/docs
- It is part of build process.

Will improve it, interested people can already check it out. 

- [x] Only create doc files for components (not import.js and index.js)


fokusferit:
For now there is still an issue with generating docs:

-> As the transpiled modules are Objects jsdocs has some issues correctly understanding it.

fokusferit:
FYI: I'm not adding this doc coverage tool for now because it is not supporting ES6 out of the box. So I will do a future PR for that process.

build/tasks/docs.js
```diff
@@ -0,0 +1,47 @@
...
+  const files = glob.sync(paths.source); // glob allows pattern matching for filenames
+  const createdMdDocs = [];
+  files
+  .filter(filename => {
+    const name = filename.match(/((\w)+(\-)*(\w))+(?=.js)/g)[0];
+    return name !== 'imports';
+  })
+  .map(filename => {
+    const splittedName = filename.split('/');
+    const filenameWithoutExt = filename.match(/((\w)+(\-)*(\w))+(?=.js)/g)[0];
```
fragsalat: You could make a constant for this regex

package.json
```diff
@@ -94,5 +98,9 @@
   "peerDependencies": {
     "preact": "7.1.0",
     "preact-compat": "3.13.1"
+  },
+  "dependencies": {
```
fragsalat: Pls remove this. We don't want react or preact as runtime dependencies.

package.json
```diff
@@ -61,11 +61,14 @@
     "chai": "3.5.0",
     "conventional-changelog": "^1.1.0",
     "css-loader": "0.26.1",
+    "documentation": "^4.0.0-beta.18",
```
fragsalat: There is no stable version?

build/tasks/build.js
```diff
@@ -1,51 +1,51 @@
...
 });
 
-gulp.task('build-css', function() {
+gulp.task('build-css', () => {
   return gulp.src(paths.css)
-    .pipe(gulp.dest(paths.output + 'es2015'))
-    .pipe(gulp.dest(paths.output + 'commonjs'))
-    .pipe(gulp.dest(paths.output + 'amd'))
-    .pipe(gulp.dest(paths.output + 'system'));
+    .pipe(gulp.dest(`${paths.output  }es2015`))
```
fokusferit: idk why this happened -> might be something with my visual code auto-fix thing.

build/tasks/docs.js
```diff
@@ -0,0 +1,47 @@
...
+const paths = require('../paths');
+const fs = require('fs-then-native');
+const glob = require('glob');
+const del = require('del');
+const vinylPaths = require('vinyl-paths');
+const docjs = require('documentation');
+
+// deletes all files in the output path
+gulp.task('clean-docs', () =>
+  gulp.src([`${paths.output}/docs/`])
```
fokusferit: 👍  you're right. Will create docs/api for that

package.json
fokusferit: 👍

package.json
fokusferit: They have a v3.0.5 which is from 2015, it looks like they have moved to v4.0.0 but didn't decided yet to introduce one stable version for that. Maybe I should open an issue ^^

PR #24, opened 2017-03-15
Date picker v3

fragsalat:
- reimplemented date-picker as React/Preact component based on flatpickr
- Flatpickr only received modifications on the locales (Kw -> CW, Mon -> Mo, Tue -> Tu, ...)
- Style is based on latest design

20170315_date-picker_ v1_my

This is how it looks in the demo_app.js

date-picker-v2 1


package.json
```diff
@@ -89,6 +91,9 @@
     "webpack-node-externals": "1.5.4",
     "wholesale-styleguide-scss": "git+ssh://git@github.com/wholesale-design-system/scss.git"
   },
+  "dependencies": {
+    "moment": "^2.0.0"
```
faboweb: you did remove momentjs as far as I understood. Is this then needed anymore?

fragsalat: Fixed. True

fragsalat:
@faboweb What do you mean with min max borders? There is nothing implemented and also not required.

faboweb:
@fragsalat With min/max-borders I mean defining a timeframe where you can pick a date from. Like: you can only select shipment between now and the end of the season. The older version had this implemented. I need this for the preorder recommendation frontend.

fragsalat:
As far as I know this is not yet part of the styleguide. Can we merge this and place this as an enhancement?

PR #26, opened 2017-03-16
Ws inline edit

alexisinwork:
Here you can see inline edit component implemented as Preact/React component.
Here how it looks as a single div block:

screen shot 2017-03-16 at 14 22 55
screen shot 2017-03-16 at 14 23 06
screen shot 2017-03-16 at 14 23 20

I am happy to have some comments to styles and logic.


alexisinwork:
@tonysaad Will also add some changes after our discussion.
He want to fix "jumps" when you change from div to input.
Waiting for his changes

tonysaad:
@alexisinwork Please check the latest changes, and let me know what do you think.
Note: There were some errors in the demo as it wasn't compatible with React Rules so I changed them a bit

src/ws-inline-edit/ws-inline-edit.js
```diff
@@ -0,0 +1,95 @@
...
+        this.editEl.focus();
+      });
+    }
+  }
+  /**
+   * Function that save text when click 'Enter' or cancel when click 'Escape' button
+   * @param {Object} e - click event
+   * @returns {Object}
+   */
+  keyAction(e) {
```
faboweb: cool

fragsalat:
@alexisinwork I like you code but there is one really small thing which @tonysaad  is missing as well. Pls put a new line between class functions^^
Beside that it looks really good :+1:

src/ws-inline-edit/ws-inline-edit.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+    this.updating(e.target.value);
+  }
+  /**
+   * Function that return value for outside use
+   * @param {Object} text - text to show
+   * @returns {Object}
+   */
+  updating(text) {
+    this.props.onUpdate(text);
```
faboweb: maybe check if the text changed?

src/ws-inline-edit/ws-inline-edit.js
```diff
@@ -0,0 +1,95 @@
+import {React, Component} from '../imports';
+require('./ws-inline-edit.scss');
```
faboweb: do we need to do this with require?

src/ws-inline-edit/ws-inline-edit.js
fragsalat: @faboweb Good point but then you would not be notified when a user erased his text.

I got another point. onUpdate is neither in the propTypes required nor you're checking if this is a function.
Everyone using the component without setting onUpdate will get an exception.

Pls reference onUpdate in the propTypes with React.PropTypes.func, define a default value () => {} and check if this is a function before calling. 

```js
static propTypes = {
  onUpdate: React.PropTypes.func
};

static defaultProps = {
  onUpdate: () => {}
}

update(text) {
  if (typeof this.props.onUpdate === 'function') {
    this.props.onUpdate(text);
  }
}
```

fragsalat:
I have another question. This component covers editable texts. How do we want to handle editable selections or dates?
Should that be in one component with a type attribute or do we want to create multiple components. Then we might rename this to something type related like `ws-editable-text`/`ws-editable-date` or `ws-inline-text-edit`/`ws-inline-select-edit`.

faboweb:
type property would work for me. but there we have maybe the date picker?

src/ws-inline-edit/ws-inline-edit.js
alexisinwork: Actually I didn't find where can I inject my styles xD
I think webpack should automay=te this bundling, but until I did it styles don't work

src/ws-inline-edit/ws-inline-edit.js
alexisinwork: Okay, thank you for mention this.
I will do it later on.

alexisinwork:
I think few components is better idea then one super big general inline edit component

src/ws-inline-edit/ws-inline-edit.js
fragsalat: Have a look at ./demo/index.scss
We don't import scss files inside the js because not every system supports this but everyone got his own build of scss files in there projects.

faboweb:
but isn't it just one property on the input field? and what about the date-picker instead of inline-edit-date?

fragsalat:
@faboweb I would create a dedicated component (ws-inline-date-edit) which inherits from date picker but renders differently and adds required functionality

alexisinwork:
@faboweb Waiting from your approvement the whole PR and then after approve you can merge it

src/ws-inline-edit/ws-inline-edit.js
```diff
@@ -0,0 +1,115 @@
...
+  static propTypes = {
+    text: React.PropTypes.string,
+    onUpdate: React.PropTypes.func
+  };
+  
+  /**
+   * @defaultProps
+   * Create default onUpdate function to prevent errors if user don't use it
+   */
+  static defaultProps = {
```
faboweb: I could not find any reference to defaultProps in the docs of Preact. Did you test, if this is working?

alexisinwork: At least no eros in console

fragsalat: https://github.com/developit/preact-render-to-string/issues/5#issuecomment-234591660

fragsalat: For me this worked as well.

faboweb: then fine :) just wanted to check

PR #36, opened 2017-03-27
Adding new variables for spacings according to the Design Sys

tonysaad:
Adding new variables for spacings according to the Design System Updates, see screenshot below:
image


faboweb:
👍

PR #42, opened 2017-03-31
Removing magic numbers from icon positioning in week-picker calendar

schluchter:


fragsalat:
@schluchter can you please create this pr against the release branch instead of master?

PR #45, opened 2017-04-07
Adopted atomic design changes

fragsalat:
Fixed components regarding to atomic design branch


schluchter:
Can we have a little presentation of the principles behind the
reorganization and the rules going forward in one of our check-ins? Thanks!

tonysaad:
👍

PR #49, opened 2017-04-12
Circleci

fragsalat:
- Added configuration for CircleCI
- Removed pre-commit hook
- Added missing dependencies


package.json
```diff
@@ -62,6 +58,7 @@
     "chai": "3.5.0",
     "conventional-changelog": "^1.1.0",
     "css-loader": "0.26.1",
+    "del": "^2.2.2",
```
faboweb: is this used?

fragsalat: Yes. It was reported as missing dependency.

https://github.com/wholesale-design-system/components/blob/61cb3b1423c164fec1018917d1d679a2d031ce70/build/tasks/clean.js

faboweb: 👍

fokusferit: Maybe use exact numbers ? Thanks to my current update nightmare issues, it might be better to stick to exact numbers in general

fragsalat: Actually I don't like to use fixed numbers because you will use outdated versions for long time in most cases. As well it appears really rarely that a new version breaks something.
But I understand that it as well makes sense and therefore I'll change it :)

PR #51, opened 2017-04-20
Freetext dropdown

fragsalat:
- Implemented Dropdown with input and button only which is used at least on table filters
- Fixed change events published by dropdown by stopping bubbling of native change events from inputs
- Fixed notification for Preact by changing string ref's to arrow functions

dropdiwn-input-only

### DIFF
--- a/src/ws-dropdown/dropdown-input.js
+++ b/src/ws-dropdown/dropdown-input.js
@@ -1,0 +1,98 @@
+import {React, Component} from '../imports';
+
+const KEY_ENTER = 13;
+
+/**
+ * Renders renders a text input inside a dropdown container to provide a free text dropdown component.
+ * Possible options are value and placeholder which are both strings.
+ */
+export class DropdownInput extends Component {
+  /**
+   * @type {Object}
+   */
+  static defaultProps = {
+    value: null,
+    placeholder: '',
+    handle: () => {}
+  };
+
+  /**
+   * @type {Object}
+   */
+  static propTypes = {
+    value: React.PropTypes.string,
+    placeholder: React.PropTypes.string,
+    handle: React.PropTypes.func
+  };
+
+  /**
+   * @param {Object} props React props
+   */
+  constructor(props) {
+    super(props);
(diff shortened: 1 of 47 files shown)


src/ws-dropdown/dropdown-input.js
```diff
@@ -0,0 +1,115 @@
...
+   * @type {Object}
+   */
+  static propTypes = {
+    value: React.PropTypes.string,
+    placeholder: React.PropTypes.string,
+    handle: React.PropTypes.func
+  };
+
+  /**
+   * @param {Object} props React props
```
fokusferit: Please add a definition about PropTypes (see WS-Header). They are not auto generated.

src/ws-dropdown/ws-dropdown.js
```diff
@@ -231,40 +245,73 @@ export class WSDropdown extends Component {
...
+        ref={element => { this.dropdownMenu = element; }}
+      />
+    );
+  }
+
+  /**
+   * Render the complete dropdown
+   * @returns {Object}
+   */
+  render() {
```
fokusferit: +1 for simplifying render

src/ws-notification/ws-notification.js
```diff
@@ -94,7 +94,7 @@ export class WSNotification extends Component {
    * @returns {void}
    */
   animateIn(notification, index) {
-    const list = this.refs.list;
```
fokusferit: you should have fix this in another smaller PR. I guess you're fixing an issue here in the "free-dropdown" PR. We should still separate stuff no matter how small it is :)

src/ws-dropdown/dropdown-input.js
fragsalat: :+1: done

src/ws-notification/ws-notification.js
fragsalat: I had to remove it to work with preact because the notification component used string refs which caused exceptions.

PR #53, opened 2017-04-26
Made components using new prop-types npm package

fragsalat:


package.json
```diff
@@ -97,6 +98,7 @@
   },
   "peerDependencies": {
     "preact": "7.1.0",
-    "preact-compat": "3.13.1"
+    "preact-compat": "3.13.1",
```
faboweb: why do we need preact-compat?

faboweb: Do I need to add preact-compat to my setup?

src/ws-dropdown/dropdown-menu.js
```diff
@@ -182,7 +182,9 @@ export class DropdownMenu extends Component {
         this.showChild(data);
         break;
       case 'change':
-        this.clearSelections();
+        if (this.context.multiple) {
```
faboweb: could you maybe add a comment what this means?

package.json
fragsalat: it is a peerDependency which means it is not installed automatically but you can install it by yourself. In general there should also be react in this list and you could choose which peer u want to fulfill.

So no you don't.

src/ws-dropdown/dropdown-menu.js
fragsalat: Will follow

PR #54, opened 2017-04-27
Fix bugs for dropdown and date picker

fragsalat:
- Added className property to dropdown items
- Added title property to dropdown items
- Fixed behavior of selected items
- Fixed handling of property changes
- Added onChange property to propagate changes in react world
- Fixed change event and styling of date picker

### DIFF
--- a/src/ws-date-picker/ws-date-picker.js
+++ b/src/ws-date-picker/ws-date-picker.js
@@ -45,7 +45,7 @@ export class WSDatePicker extends Component {
   }
 
   /**
-   * Initialize the flatpickr with given options
+   * Initialize the flatpickr with given options and prevent default change event
    * @returns {void}
    */
   componentDidMount() {
@@ -56,6 +56,8 @@ export class WSDatePicker extends Component {
       ...this.props.options,
       onChange: this.onChange.bind(this)
     });
+    // Prevent default change event bubbling
+    this.input.addEventListener('change', event => event.stopPropagation());
   }
 
   /**
@@ -74,11 +76,12 @@ export class WSDatePicker extends Component {
   }
 
   /**
-   * Destroy the flatpickr and all it's events and bindings
+   * Destroy the flatpickr and all events and bindings
    * @returns {void}
    */
   componentWillUnmount() {
     this.flatpickr.destroy();
+    this.input.removeEventListener('change', event => event.stopPropagation());
   }
 
   /**
@@ -111,7 +114,6 @@ export class WSDatePicker extends Component {
             defaultValue={this.state.value}
             placeholder={this.props.placeholder}
             ref={element => { this.input = element; }}
-            onChange={event => event.stopPropagation()}
             key="input"
           />,
           <span className="icon icon-calendar icon16" key="icon" />
--- a/src/ws-date-picker/ws-date-picker.js
+++ b/src/ws-date-picker/ws-date-picker.js
@@ -92,7 +92,7 @@ export class WSDatePicker extends Component {
    */
   onChange([selectedDate], value) {
     this.setState({value});
-    this.element.dispatchEvent(new CustomEvent('change', {detail: selectedDate, bubbles: true}));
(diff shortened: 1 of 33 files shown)


src/ws-date-picker/ws-date-picker.js
```diff
@@ -89,7 +92,7 @@ export class WSDatePicker extends Component {
    */
   onChange([selectedDate], value) {
     this.setState({value});
-    this.element.dispatchEvent(new CustomEvent('change', {detail: selectedDate}));
+    this.element.dispatchEvent(new CustomEvent('change', {detail: {date: selectedDate, value}, bubbles: true}));
     // Propagate if wanted
     if (this.props.onChange) {
       this.props.onChange(selectedDate);
```
fokusferit: It looks like the properties of ws-date-picker are not created, could you ? Or at least we should track the lack of documentation in an Issue

src/ws-date-picker/ws-date-picker.scss
```diff
@@ -263,7 +264,7 @@
     font-size: rem(14px);
     width: rem(32px);
     height: rem(32px);
-    line-height: rem(32px);
+    line-height: rem(28px); // Numbers on acumin-pro font are lower
```
fokusferit: A variable here or some calculation might be better I think

src/ws-dropdown/dropdown-menu-item.js
```diff
@@ -89,9 +93,16 @@ export class DropdownMenuItem extends Component {
       this.props.handle('show-child', this.menu);
     } else {
       if (!this.context.multiple) {
-        this.state.selected = true;
-        this.props.handle('change', this.state);
+        // If it is selected we publish null because it will be deselected in the upper menu
+        if (this.state.selected) {
+          this.props.handle('change', null);
```
fokusferit: Two lines below you do `this.state.stored = true;` do we need to reset it here maybe ? idk what 'stored' exactly means, just asking to understand it properly :D

src/ws-dropdown/dropdown-menu.js
```diff
@@ -318,8 +340,8 @@ export class DropdownMenu extends Component {
           />,
           <li className="dropdown-item-separator" key="parent-separator" />
         ]}
-        {(this.state.value && this.state.value.length) ? [
-          this.state.value.map((item, index) =>
+        {(this.state.value && this.state.value.length && (this.context.multiple || this.props.filterable)) ? [
```
fokusferit: This looks like complicated and hard to read. Maybe refactoring it into a function like "renderXyZ()" so it is easier to understand parts of the render function more easily ?

src/ws-dropdown/ws-dropdown.js
```diff
@@ -37,7 +39,8 @@ export class WSDropdown extends Component {
     limit: 10,
     orientation: 'left',
     placeholder: '',
-    value: null
+    value: null,
+    onChange: () => {}
```
fokusferit: We need a jsDoc for that! :)

src/ws-date-picker/ws-date-picker.scss
fragsalat: If you tell me how to calculate the line height for this specific font I gonna use it ;) :D

src/ws-date-picker/ws-date-picker.js
fragsalat: Actually I don't really got what you mean with the properties are not created.

src/ws-dropdown/dropdown-menu-item.js
fragsalat: Nope. I'm pretty sure that we don't need to do it. The flag stored is an indicator that the item is selected to be presented as a value or stored in values list.

src/ws-dropdown/ws-dropdown.js
fragsalat: Fixed

tonysaad:
👍 I've tested it and it works well I will add something later for 'filterable' dropdowns that the input field gets auto-focused when the dropdown opens

PR #58, opened 2017-05-09
Testing strategy for react components

fokusferit:
Adding tests for our (p)react components. 

Generally working with Jasmine is fine except that spying / mocking functionality within a component is not perfect, e.g you have to render the component first, get the component Instance (preact is returning DOM whereas React is returning the Component) and afterwards spy on "component.__proto__". Works fine currently, but the moment __proto__ changes from Preact / React changes sth these kind of tests will fail. Instead of Mocking a function I could also mock the backend Calls but introducing Sinon, Express etc. is a overkill for just testing ws-header.


package.json
```diff
@@ -71,7 +73,13 @@
...
     "gulp-eslint": "3.0.1",
+    "jasmine": "2.5.3",
     "karma": "1.6.0",
+    "karma-chrome-launcher": "2.0.0",
+    "karma-jasmine": "1.1.0",
+    "karma-mocha-reporter": "2.2.3",
+    "karma-sourcemap-loader": "0.3.7",
+    "karma-webpack": "2.0.3",
     "lite-server": "2.3.0",
     "mocha": "3.2.0",
```
fragsalat: Can we remove this?

tests/index.scss
```diff
@@ -0,0 +1,13 @@
+$use-embedded-font: true;
```
fragsalat: Why do we need styles for tests?^^

tests/ws-header/ws-header.spec.js
```diff
@@ -0,0 +1,179 @@
...
+    email: 'chuck.norris@zalando.de'
+  };
+
+  const defaultState = {
+    availableLanguages: ['de', 'en']
+  };
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    document.querySelector('body').appendChild(container);
```
fragsalat: Doe you really need to append it to the body?

tests/index.scss
fokusferit: This was just for convenience as I had to debug stuff directly in Karma Debugger, once you do that some CSS helps here.

package.json
fokusferit: I'm using all of them, e.g. here https://github.com/wholesale-design-system/components/pull/58/files#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfbR33 .

I'm using webpack etc. to build the application and run it in chrome. mocha-reporter because I like the output more then the default jasmine stuff.

tests/index.scss
fokusferit: This was just for convenience as I had to debug stuff directly in Karma Debugger, once you do that some CSS helps here.

tests/ws-header/ws-header.spec.js
fokusferit: No we don't as I said, once you start to debug stuff in Karma then it is easier to see the current component, otherwise the screen is just empty... but we can discuss this maybe with others @faboweb @tonysaad ??

tonysaad:
@fokusferit I tried running it locally and I get the following error when I run `npm run test`
`var urlAtStart = window.location.href;
                                ^
ReferenceError: window is not defined`

fragsalat:
Run gulp lint
The npm scrips are old relicts

tests/ws-header/ws-header.spec.js
faboweb: Having the ability to debug your tests is a plus.

tests/index.scss
faboweb: I would also get rid of the styles. effectively removing one layer of complexity for the tests. For unit tests, we don't need them.
If we would do screenshot tests, this is a different thing.

tests/ws-header/ws-header.spec.js
faboweb: Addition: If you never attach the elements to the DOM, you don't know if it actually doesn't fail when you do in production.

fragsalat: I think if you have to debug something you can add it temporarily to the body^^

PR #60, opened 2017-05-10
Tile chart: Simplify the properties and allow tile styling using css

inanda:
Tile Chart: Simplify properties and allow tile CSS styling per group

- Simplify the data property passed to the component
- Add the group name as a class to the tile in order to allow styling using CSS
- Add a default tile color to the CSS
- Change Demo page to show how to style the tiles per group using CSS instead of property

### DIFF
--- a/demo/demo_app.js
+++ b/demo/demo_app.js
@@ -78,8 +78,7 @@ render(
         </tbody>
       </table>
       <br />
-        <WSTilesChart title="Chart title" data={dashboardMockData.data} width={100} height={100}
-        />
+        <WSTilesChart title="Chart title" data={dashboardMockData.data} width={100} height={100} />
     </div>
   </div>
 , document.querySelector('#app-holder'));
--- a/demo/demo_app.js
+++ b/demo/demo_app.js
@@ -78,8 +78,7 @@ render(
         </tbody>
       </table>
       <br />
-        <WSTilesChart
-          title="Chart title" data={dashboardMockData.data} config={dashboardMockData.config} width={100} height={100}
+        <WSTilesChart title="Chart title" data={dashboardMockData.data} width={100} height={100}
         />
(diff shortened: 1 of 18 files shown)


demo/demo_app.js
```diff
@@ -78,8 +78,7 @@ render(
         </tbody>
       </table>
       <br />
-        <WSTilesChart
-          title="Chart title" data={dashboardMockData.data} config={dashboardMockData.config} width={100} height={100}
+        <WSTilesChart title="Chart title" data={dashboardMockData.data} width={100} height={100}
         />
```
fragsalat: You still  don't need to wrap here^^

demo/index.scss
```diff
@@ -11,3 +11,14 @@ $use-embedded-font: true;
 .container {
   margin: 20px;
 }
+
+
+//Example of how to style ws-tiles-chart using classes (.tile.groupName).
+.ws-tiles-chart .tile {
```
fragsalat: Could you pls create a own scss file for your component?

inanda: The component contains its own SCSS file. This is just some CSS for the demo page to show that you can change the color of the tiles in two ways, 
1) Sending the config properties to the component
Like:   config: {
    groupName1:'#FF4D4D',
    groupName2: '#ffcc33'
  }
 2)using .tile.groupName in the SCSS of your application

If no SCSS is specified for the tile color and no config was sent to the component, by default, it will show the tiles as green for all the groups.

demo/demo_app.js
inanda: Fixed

demo/index.scss
inanda: Ah did you mean create a own SCSS demo file for the component and import it into index.scss? I got it :)

PR #61, opened 2017-05-11
Created test component class to simplify tests

fragsalat:
I did a pretty simple but usefull enhancement by simplifying the creation of test components.

### DIFF
--- a/tests/test-component.js
+++ b/tests/test-component.js
@@ -1,0 +1,36 @@
+// eslint-disable-next-line import/extensions
+import {render} from 'imports';
+
+/**
+ * Creates and wraps React / Preact components for easier testing
+ */
+export class TestComponent {
+
+  /**
+   * Initialize the component and render the jsx as React / Preact component
+   * @param {Object} jsx JSX to render
+   */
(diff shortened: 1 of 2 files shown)


tests/test-component.js
```diff
@@ -0,0 +1,36 @@
...
+    // Preact returns HTMLDomElement with component in ._component variable || React is returning WSHeader Object
+    const result = render(jsx, this.container);
+    // eslint-disable-next-line no-underscore-dangle
+    this.component = result._component || result;
+  }
+
+  /**
+   * Helper function for querySelector and querySelectorAll
+   * @param {String} selector CSS Selector string
+   * @returns {null|Element|Array<Element>} Returns null if nothing found, Element if one was found and otherwise an Array
```
fokusferit: do you think it is better to return null, Element and Array<Element> or just null, Array<Element> ?

On the one hand I like the idea but I could imagine that you get mixed up easily.

fokusferit:
generally I like the idea and besides the one comment I will merge it on monday when I'm back in office. I really like the idea of having the TestComponent 👍

Issue #47, opened 2017-04-11
Usage section of README


qlibin:
Usage section of README says:
> run `gulp` for building / watcher
> `gulp test` for test / watcher 

_I think it's explicitly implied that I have to run also `npm install` (which in my opinion should also be documented)_

After doing following:
```
$ npm install
$ gulp
```

`gulp` fails with following output:
```
[14:42:39] Using gulpfile ~/dev/js/components/gulpfile.js
[14:42:39] Task 'default' is not in your gulpfile
[14:42:39] Please check the documentation for proper gulpfile formatting
```

`gulp test` fails with following:
```
$ gulp test
[14:56:16] Using gulpfile ~/dev/js/components/gulpfile.js
[14:56:16] Starting 'test'...
11 04 2017 14:56:16.849:ERROR [config]: File /Users/alibin/dev/js/components/build/tasks/../../karma.conf.js does not exist!
```

fragsalat:
Either you run `gulp build` to build the components or to run the demo page with react `npm run start` or with preact `npm run start:preact`

qlibin:
Thank you @fragsalat but this doesn't help :(

Both `npm run start` and `npm run start:preact` show the error:
```
ERROR in ./~/css-loader!./~/sass-loader/lib/loader.js!./demo/index.scss
Module build failed: 
      margin-top: $space-l;
                 ^
      Undefined variable: "$space-l".
      in /Users/alibin/dev/js/components/src/ws-header/ws-header.scss (line 230, column 19)
 @ ./demo/index.scss 4:14-117
 @ ./demo/demo_app.js
 @ multi (webpack)-dev-server/client?http://localhost:8080 ./demo/demo_app.js
webpack: Failed to compile.
```

:man_shrugging: 

FYI: my node version is v7.7.4 and npm 4.1.2

fragsalat:
@qlibin Can you pls run `npm i -D https://github.com/wholesale-design-system/scss.git`

I guess the latest master of scss repo isn't installed.

qlibin:
There was one warning during the `npm i -D https://git...`:
```
npm WARN url-loader@0.5.7 requires a peer of file-loader@* but none was installed.
```

But it helped! Thank you

Issue #46, opened 2017-04-11
Documentation is not clear regarding how to use Demo


qlibin:
It says:

> In 'demo/index.html' you will find already build web-components. And see how you can integrate them in a vanilla setup.

But all that `index.html` has inside is basically import of one non existing script:
```
    <script src="bundle.js"></script>
```

fragsalat:
True, The built components are inside the dist folder and demo should use them

faboweb:
`npm run serve` is also not really working on master -> components produce errors

PR #63, opened 2017-05-23
fix(documentation): Update readme

fokusferit:
fix(documentation): Update readme 

Should fix the following issues:

#47 #46 #9


README.MD
```diff
@@ -1,97 +1,60 @@
...
+## Getting started (How to integrate)
 
-## How to extend
+The package is not yet published on npm as we are still working in a beta phase. 
 
-Create a folder under `webcomponents` with the name of your web-component. In there you can have .html, .scss and .js files that will be merged into one file by the build process.
+First of all you should install react / preact (if not already done) and our scss rules:
 
-### How do Web-Components work
+- `npm i -D -E preact prop-types` or `npm i -D -E react react-dom prop-types`
```
fragsalat: preact-compat or react-dom is missing I would say or I'm wrong?

faboweb:
Are the docs generated or did you wrote them? If they are generated, can we remove them from the repo?

Awesome work.

README.MD
fokusferit: I don't think for our components. It is a difficult question because if someone wants to use preact + our components but is using react for it's own project then they need to decide to use preact and drop react completely. And for non-react stacks there shouldn't be a need for preact-compat as we are assuming that preact is kind of a subset of react. That's why we test for both frameworks. 

But maybe I'm missing sth completely ? @faboweb any thoughts ? :)

fragsalat: Maybe a stupid question but why should someone load an extra library which behaves the same like reakt if react is already loaded?^^ So I gues no one will have an React app and plan to just use the components with preact^^

PR #64, opened 2017-05-24
fix(header): Fixing a bug in Authentication URL that it had newlines

tonysaad:
Fixing a bug in Authentication URL that it had newlines which Omer reported


okarahan:
Looks good

okarahan:
:+1:

Issue #67, opened 2017-05-30
Tile Chart does not allow mouse events


inanda:
Tile chart does not accept mouse event functions.

- Add the possibility to call a function onMouseEnter and onMouseLeave.

- In addition, we also need to call a function by clicking to the tile and we need to know the tile identifier and the group of the tile that was clicked.

PR #68, opened 2017-05-30
feat(Tile chart) - mouse events

inanda:
Pull request for issue: #67

Description (short):  
- Added onMouseEnter and onMouseLeave events to the chart
- Added onCLick event to the tile


src/ws-tiles-chart/ws-tiles-chart.js
```diff
@@ -24,7 +29,10 @@ export class WSTilesChart extends Component {
     maxTileSize: 25,
     minTileSize: 8,
     width: 80,
-    height: 80
+    height: 80,
```
fokusferit: just a question, because I was not involved in the Design talks, are these default values from UX or did you choose some on your own ?

src/ws-tiles-chart/ws-tiles-chart.js
```diff
@@ -10,7 +10,12 @@ import {Tile} from './tile';
  * @property {Object} props.config Defines the color of each group of tiles
  * @property {string} props.title Title of the chart
  * @property {number} props.maxTileSize Defines the maximum size that the tile can be
```
fokusferit: Is this maxTIleSize (also min...) in Pixel? I would recommend that for these two you also add "(in pixel)" at the end of the documentation so it is absolutely clear 😄

src/ws-tiles-chart/ws-tiles-chart.js
```diff
@@ -64,20 +75,26 @@ export class WSTilesChart extends Component {
    * @returns {number}
    */
   getTileSize() {
-    const groups = this.props.data.groups || {};
+    const {height, width, maxTileSize, minTileSize, data} = this.props;
```
fokusferit: +1

src/ws-tiles-chart/ws-tiles-chart.js
inanda: I was not involved in the Design talks. This is a component that the UX designed for one of our applications and I implemented while creating it. I added the default size on my own, I did not ask it from him. I think we will need it in different sizes for our application. But maybe we could ask a default from them.

src/ws-tiles-chart/ws-tiles-chart.js
inanda: +1 Done!

fragsalat:
Don't forget to make a PR from release branch to master after merging this.

PR #75, opened 2017-06-02
Drop down support for disabled property

jseb:
Pull request for issue: #75

Are there more related issue(s): # #

Component Screenshot:

Description (short):

### DIFF
--- a/src/ws-dropdown/ws-dropdown.js
+++ b/src/ws-dropdown/ws-dropdown.js
@@ -303,10 +303,6 @@ export class WSDropdown extends Component {
     item.classList.add(animationClass);
   }
 
-  disabledStyle() {
-    return this.props.disabled ? ' disabled' : '';
-  }
-
   /**
    * Renders the dropdown trigger element
    * @returns {Object}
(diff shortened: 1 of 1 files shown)


src/ws-dropdown/ws-dropdown.js
```diff
@@ -301,6 +303,10 @@ export class WSDropdown extends Component {
     item.classList.add(animationClass);
   }
 
+  disabledStyle() {
+    return this.props.disabled ? ' disabled' : '';
```
fragsalat: We aligned on a naming convention which says that states are prefixed with an `is-` and modifies with an `mod-`
So could you please rename that class to an `is-disabled`?

As well it would be cool if you would create a variable inside the render function since this function is used only there and we don't need to blow up the function amount.

jseb: Sure, sounds good, I will make the changes

src/ws-dropdown/ws-dropdown.js
```diff
@@ -312,14 +318,26 @@ export class WSDropdown extends Component {
     }
     switch (this.props.type) {
       case 'anchor':
-        return <a className="dropdown-trigger" onClick={() => this.open()}>{icon} {this.state.text}</a>;
+        return <a className={"dropdown-trigger" + this.disabledStyle()}
```
fragsalat: I guess the linter will complain here that you have to use a template string instead. So it would finally look like this
```js
return <a
  className={`dropdown-trigger ${disabledStyle}`}
  onClick={() => this.open()}
>
  {icon} {this.state.text}
</a>;
```

src/ws-dropdown/ws-dropdown.js
```diff
@@ -312,14 +318,26 @@ export class WSDropdown extends Component {
...
-        return <div className="dropdown-trigger select-box" onClick={() => this.open()}>{icon} {this.state.text}</div>;
+        return <div className={"dropdown-trigger select-box" + this.disabledStyle()}
+          onClick={() => this.open()}>
+            {icon} {this.state.text}
+        </div>;
       case 'icon':
       default:
-        return <a className="dropdown-trigger" onClick={() => this.open()}>{icon}</a>;
+        return <a className={"dropdown-trigger" + this.disabledStyle()}
+          onClick={() => this.open()}>
```
fragsalat: Could you pls execute gulp lint If you didn't already? I'm not sure if the formatting is correct

Issue #81, opened 2017-06-06
Automatically generate API docs


fragsalat:
In order to build a good styleguide documentation page we want to automatically generate markdown files for our components API.
This should be done on every branch build  / merge to master automatically.

Issue #62, opened 2017-05-23
Update documentation.js


fokusferit:
With the release of beta19 of documentationjs they removed the buildSync functionality which is now breaking our generate task for docs. https://github.com/documentationjs/documentation/blob/master/CHANGELOG.md#400-beta19-2017-04-10 

We should update to one of the rc versions and update the gulp tasks accordingly.

fokusferit:
Is part of #81

fokusferit:
Looks like done. 
Update cycle is needed anyway so closing this specific one.

PR #82, opened 2017-06-06
Doc generation

fragsalat:
Pull request for issue: #81 #62 

Description:
This pr contains fixed generation of API documentation and integration into the circle ci build process.


build/tasks/docs.js
```diff
@@ -6,45 +6,57 @@ const glob = require('glob');
 const del = require('del');
 const vinylPaths = require('vinyl-paths');
 const docjs = require('documentation');
+const REGEX_FILE_NAME = /((\w)+(-)*(\w))+(?=.js)/g;
```
faboweb: just a comment: I like REGEX too, but they are damn hard to read.

build/tasks/docs.js
```diff
@@ -6,45 +6,57 @@ const glob = require('glob');
...
-    const folder = `${paths.docsOutput}${mdDoc.folder}`;
-    const filePath = mdDoc.file !== 'index' ? `${paths.docsOutput}${mdDoc.folder}/${mdDoc.file}.md` : `${paths.docsOutput}${mdDoc.file}.md`;
-    if (!fs.existsSync(folder)) {
-      fs.mkdirSync(folder);
-    }
-    return fs.writeFile(filePath, mdDoc.markdown);
-  });
-  return Promise.all(writeMdFiles);
+gulp.task('build-jsdoc-to-md', done => {
+  Promise.resolve(glob.sync(paths.source))
```
faboweb: nice chain 👍

build/tasks/docs.js
```diff
@@ -6,45 +6,57 @@ const glob = require('glob');
...
+      groups[doc.folder] = groups[doc.folder] || [];
+      groups[doc.folder].push(doc);
+      return groups;
+    }, {}))
+    // Create markdown files for each folder
+    .then(groups => Object.keys(groups).forEach(folderName => {
+      const group = groups[folderName];
+      const fileName = `${paths.apiDoc}/${folderName}.md`;
+      const normalize = doc => doc.markdown.replace(/<!--.*?-->\n*/g, '');
+      const content = `# ${folderName}\n${group.map(normalize).join('')}`;
```
faboweb: why the hash?

build/tasks/lint.js
```diff
@@ -1,9 +1,25 @@
 var gulp = require('gulp');
 var paths = require('../paths');
 var eslint = require('gulp-eslint');
+const glob = require('glob');
+const docjs = require('documentation');
+const REGEX_FILE_NAME = /((\w)+(-)*(\w))+(?=.js)/g;
```
faboweb: maybe give an acceptable example

build/tasks/lint.js
```diff
@@ -1,9 +1,25 @@
...
 
 // runs eslint on all .js files
 gulp.task('lint', function() {
+  Promise.resolve(glob.sync(paths.source))
+    // Filter file names to match tag name pattern
+    .then(fileNames => fileNames.filter(fileName => {
+      const name = fileName.match(REGEX_FILE_NAME)[0];
+      return name !== 'imports';
```
faboweb: what are the "imports" files and why are they ignored?

build/tasks/lint.js
```diff
@@ -1,9 +1,25 @@
...
+    }))
+    // Generate markdown for each file
+    .then(fileNames => fileNames.forEach(fileName =>
+      docjs.lint(fileName, {}).then(lintOutput => {
```
faboweb: is this generating markdown? it says lint?!

build/tasks/docs.js
fragsalat: I just moved them up to a global constant. This regex seem to match a file name pattern like this ws-something-somehting.js
Maybe it could be done simpler but there was no need to simplify it right now in my eyes^^

doc/components/dropdown.md
```diff
@@ -0,0 +1,138 @@
+# Dropdown
```
faboweb: I would have put new docs in a separate PR

src/ws-dropdown/ws-dropdown.js
```diff
@@ -229,7 +229,7 @@ export class WSDropdown extends Component {
       itemsToWrap = items ? [items] : [];
     }
     return itemsToWrap.map(item => {
-      const enriched = typeof item === 'object' ? item : {label: item};
+      const enriched = typeof item === 'object' ? item : {label: item, value: item};
```
faboweb: the same with small fixes :)

build/tasks/docs.js
fokusferit: I checked it again, as it was my regex (feels over-engineered :D).

Could you just try out this one: ` /((-)*(\w))+(?=.js)/g`instead of the one above. What it basically does is, it 'captures' all words of a filename until ".js". It also checks for a beginning dash of the name so it can capture the whole filename. If we would just use "\w (word)" it would only catch the first ws and stop.

We could write some code but this would be:

- Split filename `split(-)`;
- Split the last Element of this Array again `split(.)` as array would be `['ws', 'input', 'edit.js']`
- create a new Array combining both
- `.join(-)` to get "ws-input-edit".

But I thought a one line regex makes sense here instead of doing this slicing, splitting thing ^^.

build/tasks/docs.js
fokusferit: I think this is related to how the API documentation is later included in the styleguide.
I remember that @fragsalat told me that his markdown-interpreter for the styleguide needs this "#" which is basically markdown headline (h1) :D

build/tasks/docs.js
```diff
@@ -6,45 +6,57 @@ const glob = require('glob');
...
+    .then(docs => docs.reduce((groups, doc) => {
+      groups[doc.folder] = groups[doc.folder] || [];
+      groups[doc.folder].push(doc);
+      return groups;
+    }, {}))
+    // Create markdown files for each folder
+    .then(groups => Object.keys(groups).forEach(folderName => {
+      const group = groups[folderName];
+      const fileName = `${paths.apiDoc}/${folderName}.md`;
+      const normalize = doc => doc.markdown.replace(/<!--.*?-->\n*/g, '');
```
fokusferit: +1 for removing docjs comment ^^

build/tasks/lint.js
fokusferit: I would say there is no need to ignore it. imports.js is where we abstract the our React / Preact imports from node_modules.

build/tasks/lint.js
fokusferit: Agreed, I wouldn't use docjs.lint for linting our files even if they use eslint under the hood.

src/index.js
```diff
@@ -5,13 +5,11 @@
  * Here we should add more descriptions
  * Below are all components referenced
  *
- * - See {@link ws-header/ws-header.md|WSHeader}
- * - See {@link ws-date-picker/ws-date-picker.md|WSDatePicker}
- * - See {@link ws-dropdown/ws-dropdown.md|WSDropdown}
- * - See {@link ws-inline-edit/ws-inline-edit.md|WSInlineEdit}
- * - See {@link ws-notification/ws-notification.md|WSNotification}
- * - See {@link ws-week-picker/ws-week-picker.md|WSWeekPicker}
- * - See {@link ws-tiles-chart/ws-tiles-chart.md|WSTilesChart}
```
fokusferit: Why did you remove them ?

build/tasks/docs.js
faboweb: we could add this as a comment

fragsalat: The hash is markdown h1 tag^^

build/tasks/lint.js
fragsalat: I will ignore not only imports but also index.js because the API doc for this is not needed in my eyes.

build/tasks/lint.js
fragsalat: Oh this is a relict and will be removed soon

src/ws-dropdown/ws-dropdown.js
fragsalat: But the fixes appeared during writing the docs^^

build/tasks/docs.js
fragsalat: It is necessary. Otherwise there would be sooo much white space

src/index.js
fragsalat: Because it would generate an empty page on the styleguide which is not needed. We need the API docs for the classes and not for a simple index file

build/tasks/lint.js
fragsalat: Wrong it is not a relict. This lint's the code for invalid annotations. +1 for moving it into a separate task

src/index.js
fokusferit: ok, never mind. I think the whole Markdown Index.md was to have a easy to use API documentation hosted on Github. As we move to show them in our styleguide anyway maybe this index.md is unnecessary anymore.

Issue #69, opened 2017-05-30
Tile chart highlight by group on mouse over


inanda:
Currently, it's not possible to highlight the tiles by group on mouse over.
The idea is to add 'group-over' class to all the tiles of the group when the mouse is over one of the tiles of the group.

inanda:
This is implemented and merged. I'll close it

PR #84, opened 2017-06-06
feat(Tile chart) - highlight group

inanda:
Pull request for issue #69 

- Add 'group-over' class to all the tiles of the group when the mouse is over one of the tiles of the group


demo/index.scss
```diff
@@ -15,10 +15,19 @@ $use-embedded-font: true;
 
 //Example of how to style ws-tiles-chart using classes (.tile.groupName).
 .ws-tiles-chart .tile {
+  &.lessThan2h.group-over {
```
faboweb: nice describtive classes

src/ws-tiles-chart/tile.js
```diff
@@ -49,9 +51,11 @@ export class Tile extends Component {
 
     return (
       <div
-        className={`tile ${this.props.groupName}`}
+        className={`tile ${groupName} ${className}`}
         style={style}
-        onClick={() => this.props.onClick(this.props.groupName, this.props.identifier)}
+        onClick={() => this.props.onClick(groupName, identifier)}
```
faboweb: what is the difference between groupName and identifier?

inanda: The tile chart contains multiple groups that can be shown in different colors (depending on your css) and each group contains multiple tiles.
- groupName identifies the group of the tile that was clicked 
- identifier identifies a single tile

Issue #30, opened 2017-03-22
Week-Picker


faboweb:
In the preorder recommendations frontend we need this component.

I already started implementing using the design aligned before for the web component.

faboweb:
#32

fokusferit:
@faboweb do you have any update on that ? maybe it is already done :) ?

fragsalat:
As far as I saw it the component isn't working and doesn't look well.

fragsalat:
BTW Where is the design for this component?

faboweb:
@fokusferit This is open as we did not have a UX-Alignment meeting in weeks.

@fragsalat If you figured sth is not working please add a detailed issue for that. Please avoid unconstructive critizism like above. I am always open to take things offline.

The design was aligned with Mehmet and Señor Schluchter before the change to Preact. We can however reopen the design discussion, if you think it dramatically differs from what you think it should look like.

faboweb:
BUG: When you click on it there is an error...

myueksel:
Updated of CW picker... aligned to new spacings and styles (same as the date picker).

Invision:
https://zalando.invisionapp.com/d/main/#/console/10454474/235119440/preview

![cw picker v1](https://cloud.githubusercontent.com/assets/25901307/26345702/fe129c0e-3fa3-11e7-8504-705f19fc3355.png)

myueksel:
please check whether there are magic numbers some where in code.

fokusferit:
I'm closing this as there was a PR which merged it. 
If there are open things, please create a new Issue

PR #85, opened 2017-06-12
Ws week picker

faboweb:
Pull request for issue: #30

Component Screenshot:
image

Description (short): 
Component was broken because of some typos. + Added new styling by UX.

### DIFF
--- a/src/ws-week-picker/ws-week-picker.scss
+++ b/src/ws-week-picker/ws-week-picker.scss
@@ -1,27 +1,30 @@
 @import "~wholesale-styleguide-scss/utils/variables";
 
-// PROPOSAL for mixin
-$padding: 1rem;
+$padding: $space-m;
 // to allow padding with setting a transparent border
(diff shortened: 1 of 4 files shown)


faboweb:
I don't know why, but there is some foreign commit in there ^^

src/ws-week-picker/ws-week-picker.scss
```diff
@@ -15,8 +16,10 @@ $padding: 1rem;
     }
   }
   @if $bottom != 0 {
-    tr:first-child:after {
-      height: $bottom;
+    tr:last-child {
+      td, th {
+        padding-bottom: $bottom !important;
```
fragsalat: Just want to not that using !important is not the best solution :)

faboweb: valid point. maybe I change it tomorrow.

myueksel:
looks great!

PR #86, opened 2017-06-16
Doc generation

fragsalat:
- Try to fix version generation by allowing circle ci to push to master
- Fixed generation of api doc
- Added documentation for date picker and notification


circle.yml
```diff
@@ -19,7 +19,7 @@ checkout:
   post:
     - git pull origin ${CIRCLE_BRANCH}
     - git clone git@github.com:wholesale-design-system/styleguide.git
-    - git config --global user.email circleci@circleci
+    - git config --global user.email t.schlage@gmx.net
```
faboweb: maybe use different PRs

faboweb: in the future

doc/components/date-picker.md
```diff
@@ -0,0 +1,45 @@
...
+## Custom options
+Since our date picker is based on the flatpickr you can configure it like the flatpickr by passing an
+object through the `options` attribute. All possible options can be found [here](https://chmln.github.io/flatpickr/options/).
+
+<ws-date-picker value="2017-06-15" format="Y-m-d" options.bind="{minDate: '2017-06-01', maxDate: '2017-06-24'}"></ws-date-picker>
+```html
+<ws-date-picker
+  value="2017-06-15"
+  format="Y-m-d" 
+  options='{"minDate": "2017-06-01", "maxDate": "2017-06-24"}'>
```
faboweb: should we change the format directly to min/max attributes like discussed?

doc/components/date-picker.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Custom options
+Since our date picker is based on the flatpickr you can configure it like the flatpickr by passing an
+object through the `options` attribute. All possible options can be found [here](https://chmln.github.io/flatpickr/options/).
+
+<ws-date-picker value="2017-06-15" format="Y-m-d" options.bind="{minDate: '2017-06-01', maxDate: '2017-06-24'}"></ws-date-picker>
+```html
+<ws-date-picker
+  value="2017-06-15"
+  format="Y-m-d"
```
faboweb: do we need format?

doc/components/dropdown.md
```diff
@@ -26,6 +26,11 @@ like `text="Click me" icon="icon-filter"`. As following you can see an example o
   </div>
 </div>
 
+**Disabled**
+As any common html element the dropdown can be disabled by adding the `disabled="1"` flag to it.
```
faboweb: disabled="1" is confusing. disabled is not really a boolean attribute: https://codepen.io/anon/pen/NgpZYe

faboweb: I recommend just using "disabled"

doc/components/notification.md
```diff
@@ -0,0 +1,52 @@
...
+you have to add the tag `<ws-notification></ws-notification>` to the most root container you can add.
+So either document body or the application root depending on your setup. This notification element is
+just the holder / list of notifications.
+
+## Creating notifications
+To create a notification you have to publish a custom event to the window containing the relevant data.
+The type of the event has to be `ws-notification-open` and the details has to be an object containing:
+- **title**: string, required
+- **description**: string, optional
+- **type**: info|error|warning|success, default: info
```
faboweb: idk maybe mention those are strings?

doc/components/dropdown.md
```diff
@@ -26,6 +26,11 @@ like `text="Click me" icon="icon-filter"`. As following you can see an example o
...
+The style of the trigger will change and the dropdown con not be opened any more.
```
faboweb: typo: con => can

doc/components/date-picker.md
fragsalat: Actually it was a requirement by CAT and it makes sense because the backend requires specific date formats.

doc/components/date-picker.md
fragsalat: Yes but lets to this in a specific PR :)

doc/components/dropdown.md
fragsalat: It is not working. I described why somewhere in the document or in the code. The problem is that disabled will result in the property disabled: "" which will be false.

doc/components/date-picker.md
faboweb: I think the displaying of the date should be unified. @wholesale-design-system/ux 
Processing the date afterwards to send it to the backend can be done by each application differently.

doc/components/date-picker.md
faboweb: ok

doc/components/dropdown.md
faboweb: is this a React thing?
if it is about implementation do the check `props.disabled!==undefined` (or sth like that) this results in the same behaviour as native

fragsalat: It is a HTML and wrapper thing. If you call getAttribute('name') and you define disabled as a flag it will return an empty string which will be false. So either you define a list of boolean attributes or create the convention that every attribute which has an empty string will be boolean. Btw as far as I know `disabled="1"` is valid HTML.

doc/components/date-picker.md
fragsalat: We can align for Wholesale but since we want to give the styleguide to others we can not enforce them to use a specific formatting.

doc/components/dropdown.md
faboweb: It is valid HTML but I just say it gives the wrong impression as `disabled="false"` is still disabling a button: https://codepen.io/anon/pen/owwdEp

My approach is here https://codepen.io/anon/pen/GEEdXP

circle.yml
fokusferit: I would say, in the future something like "tech-fabric@zalando.de" where all engineers working with it are subscribed ^^

doc/components/dropdown.md
fokusferit: Well, as you both discussed this I checked if `disabled="1"` is really valid XHTML specification and it is not. If you use W3C Validator it shows you an error.

> For example, setting the disabled attribute to the value "false" is disallowed, because despite the appearance of meaning that the element is enabled, it in fact means that the element is disabled (what matters for implementations is the presence of the attribute, not its value).

or 

> A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

from here:
http://w3c.github.io/html/single-page.html#sec-boolean-attributes 

so basically:

```
<input type="text" disabled />
<input type="text" disabled="" />
<input type="text" disabled="disabled" />
```

See: https://codepen.io/fokusferit/pen/gRoWKd (disabled="{0,1, lala, ...}" everything disables the button.

doc/components/date-picker.md
faboweb: discussed in #87

circle.yml
fragsalat: As I said you can create such a user if you want :D
I don't know where to store the credentials in a secure way so that also peoples after us get the access.

PR #90, opened 2017-06-30
Ws week picker preact

faboweb:
Pull request for issue: #85 

Description (short): Added documentation and tests


tests/async.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Helper that returns a promise which resolves after 10ms or the time passed as an argument
+ * @param {number} timeout time to wait until resolving promise
+ * @returns {Promise}
+ */
+export function wait(timeout = 10) {
+  const promise = new Promise(resolve => {
+    setTimeout(() => {
```
fragsalat: Even if I don't understand why this function is necessary I just want to give the hint for a simpler written code
```js
export function wait(timeout = 10) {
  return new Promise(resolve => setTimeout(resolve, timeout));
}
```

Issue #93, opened 2017-07-07
WSDropdown does not set the label of the preselected value


inanda:
The following code, preselects "New" as value. 
It's possible to see it by clicking on the dropdown, but it still  #shows "Select" as the label.

` <WSDropdown text="SELECT..." type="select" multiple={false} items={[
        'New',
        'Open'
      ]} value={"New"}/>`

![screen shot 2017-07-07 at 17 00 57](https://user-images.githubusercontent.com/682619/27963636-eea825ee-6335-11e7-8340-20da5cf69454.png)

PR #94, opened 2017-07-10
fix(dropdown): Show correct label for preselected value

inanda:
Pull request for issue: #93
- Show correct label for preselected value

### DIFF
--- a/src/ws-dropdown/ws-dropdown.js
+++ b/src/ws-dropdown/ws-dropdown.js
@@ -140,7 +140,7 @@ export class WSDropdown extends Component {
    * @returns {String}
    */
   getTextFromValue(value) {
-    let text = (this.state) ? this.state.text : '';
+    let text = this.state ? this.state.text : '';
     // Check if we have to update the text value
     if (this.props.type === 'select') {
       if (Array.isArray(value)) {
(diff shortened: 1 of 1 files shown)


src/ws-dropdown/ws-dropdown.js
```diff
@@ -140,11 +140,11 @@ export class WSDropdown extends Component {
    * @returns {String}
    */
   getTextFromValue(value) {
-    let text = this.state.text;
+    let text = (this.state) ? this.state.text : '';
```
fragsalat: You don't need the surrounding ()

src/ws-dropdown/ws-dropdown.js
```diff
@@ -181,8 +181,11 @@ export class WSDropdown extends Component {
    * @returns {Object}
    */
   createState(props) {
+    const isEmptyValue =
+      !props.value || Array.isArray(props.value) && props.value.length === 0;
+
     const state = {
-      text: props.text || this.getTextFromValue(props.value),
+      text: (!isEmptyValue) ? this.getTextFromValue(props.value) : props.text,
```
fragsalat: Again you can omit the ()

src/ws-dropdown/ws-dropdown.js
faboweb: this.state should never be empty, as it's assigned in the constructor. in which situation did this raise an error?

inanda: See my changes in the other function. getTextFromValue is now called from createState in order to set the text value. That's where it would throw an error.

PR #97, opened 2017-07-18
fix(dropdown): Made dropdown as wide as select is

fragsalat:
If you select the dropdown type select the dropdown container will have the same width as the select element

### DIFF
--- a/demo/demo_app.js
+++ b/demo/demo_app.js
@@ -19,7 +19,7 @@ render(
       <WSWeekPicker onChange={({year, week}) => console.log('New week selected:', week)} />
       <br />
       <br />
-      <WSDropdown text="Multiple" type="select" placeholder="Filter values.." filterable multiple items={[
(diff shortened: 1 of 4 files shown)


schluchter:
Suggestion: Can we make it configurable to use a max-width, and default to the width of the parent? The width of the parent is arbitrary, but the length of the data that we need to display is usually well-known in advance. With configurability, we can make it easier for implementers to make the right decision for the user.

fragsalat:
@schluchter This is a new issue and requirement, could you please create a issue for this?

tonysaad:
@fragsalat As that branch is merged, but the files aren't built on `dist/` folder yet.
Is it something that should run automatically via CircleCI or do I have to buid them manually ?

*Asking because now the `$heading-line-height` doesn't exist anymore in `scss` so the component is using an undefined variable which causing an error.

PR #99, opened 2017-07-18
feat(header): Implemented new header with authentication, locale mana…

fragsalat:
Implemented the header according to design in invision with locale management and authorization.
- Created documentation about usage of the component
- Added some simple unit tests for authorization (more needed)
- Re-Implemented header

Here is the documentation: https://wholesale-design-system.github.io/styleguide/components/ws-header/#/components/header


src/ws-header/simple-steam.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * This utility class implements a simple stream to listen continuously for changes
+ */
+export class SimpleSteam {
```
tonysaad: I can't get my hands fully around this simple stream, I see you use it for listening to changes in localstorage but not sure why is that.
* Rename it to "Stream" not "Steam" heheheh

src/ws-header/storage/cookie-storage.js
```diff
@@ -0,0 +1,66 @@
+import {AbstractStorage} from './abstract-storage';
```
tonysaad: Are those classes only for mocking Cookies and Localstorage for the tests ?

src/ws-header/authorization.js
```diff
@@ -0,0 +1,170 @@
...
+   * @param {Object} params Response parameters containing access and refresh token
+   * @returns {void}
+   * @private
+   */
+  updateTokens(params) {
+    const expires = params.expires_in ? parseInt(params.expires_in, 10) : 3600;
+    this.storage.set('access_token', params.access_token);
+    this.storage.set('refresh_token', params.refresh_token);
+    this.storage.set('expires_at', new Date().getTime() + expires * 1000);
+    // Put data into authorized stream
```
tonysaad: I see her it's pushed to the stream, Would it be accessible to the user application to access that stream ?

src/ws-header/authorization.js
```diff
@@ -0,0 +1,170 @@
...
+    // Check if the access token will expire soon and we can refresh it
+    this.checkExpiration();
+  }
+
+  /**
+   * Refresh access token if it is expired and a refresh token is available
+   * @returns {void}
+   * @private
+   */
+  checkExpiration() {
```
tonysaad: 👍

demo/index.scss
```diff
@@ -1,4 +1,5 @@
 $use-embedded-font: true;
+$schema: 'water-blue';
```
tonysaad: 💯

src/ws-header/storage/cookie-storage.js
fragsalat: This classes are to abstract the platform access to localStorage and Cookies. This makes code cleaner, better testable and more self explanatory

src/ws-header/simple-steam.js
fragsalat: Thx for the hint :)
I had the problem that multiple parts in the authorization can change the access token multiple times. So there were two ways. Either I'll pass a callback to the constructor header or I'll go with a reactive stream approach. At the point where I implemented this the code looked a bit different and this was the most attractive way for me. Right now I probably would go with the callback approach since it is much simpler to understand for others.

Issue #101, opened 2017-07-20
Tileschart does not update tile size on update


inanda:
When the props updates and the component serenaders, the size of the tiles has to be updated.

Currently, it does not, because we calculated the tile size only on componentWillMount.

inanda:
This issue is fixed. I'll close it.

PR #102, opened 2017-07-20
fix(tiles-chart): update tile size when the component updates

inanda:
Pull request for issue: #101

- Calculate the tile size also on componentWillReceiveProps, so that it updates when the properties of the component changes and it renders again.


antnascimento:
:+1:

PR #112, opened 2017-08-08
User Testing Results: UX Improvements of the Header

tonysaad:
screen shot 2017-08-08 at 15 49 42

UX Improvements of the Header according to latest User Testing session cc. @myueksel 
- Link color to be $miroschka
- Font Weight to be Bold
- Current State style
- Hide Leve-2 on click


src/ws-header/ws-header.scss
```diff
@@ -65,6 +69,15 @@ $header-border-color: schema-color(4) !default;
   }
 }
 
+@mixin list-item() {
+  font-weight: bold;
+  color: $miroschka;
+  &.is-current {
```
fragsalat: is-active would be a more standard name since we use it for every active state :)

src/ws-header/ws-header.js
```diff
@@ -288,13 +289,14 @@ export class WSHeader extends Component {
...
           onMouseLeave={() => this.leaveLevel2()}
+          onClick={() => this.leaveLevel2()}
           ref={element => { this.level2 = element; }}
         >
           {this.props.links.map((parent, index) =>
             parent.children && parent.children.length &&
             <ul className="main-sub-menu" key={`sub-menu${index}`} ref={element => { this.subMenus[index] = element; }}>
               {parent.children.map((child, childIndex) =>
-                <li key={`sub-link-${index}-${childIndex}`}>
+                <li key={`sub-link-${index}-${childIndex}`} className={(child.isCurrent) ? 'is-current' : null}>
```
fragsalat: active: true would be enough I think :)

src/ws-header/ws-header.scss
tonysaad: Yeah that's right, didn't wanna use it coz you use it for the menu itself

src/ws-header/ws-header.js
tonysaad: What do you mean ?

fragsalat: This just about the naming. The menu item is active and we use is-active everywhere as active state.

fokusferit:
@tonysaad any progress ? As I was on vacation and this is still open :) 
We shouldn't rely on schlage only here ^^

Issue #119, opened 2017-08-29
Demo page doesn't work in Preact mode


szafranek:
When running a demo page via `npm run start:preact` the server starts, but the [demo page](http://localhost:8080/) fails to load components due to a JavaScript error.

React mode (`npm run start`) works fine.

PR #120, opened 2017-08-29
Make the Demo Work Great Again in Preact Mode

szafranek:
Pull request for issue: #119 

This PR will:
* Upgrade Preact and `preact-compat` to latest stable versions. That will make the demo page run again (#119).
* Fix linting problems in scripts used by the demo page.
* Change the title of the demo page, so it doesn't reference Preact anymore. The demo works just as well with React.
* Remove `peerDependencies` section from `package.json`. All deps listed there are already included in the `dependencies` section.
* Remove fallback for `PureComponent`, since it's now [supported](https://github.com/developit/preact-compat/issues/139) by `preact-compat`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,8 +84,8 @@
     "mocha-webpack": "0.7.0",
     "node-sass": "4.5.2",
     "pre-commit": "1.2.2",
-    "preact": "7.2.1",
-    "preact-compat": "3.14.3",
+    "preact": "8.2.5",
+    "preact-compat": "3.17.0",
     "prop-types": "15.5.8",
     "react": "15.5.4",
     "react-dom": "15.5.4",
@@ -101,11 +101,5 @@
     "webpack-node-externals": "1.5.4",
     "fabric-scss": "git+https://github.com/fabric-design/scss.git#master",
     "yargs": "7.0.2"
-  },
-  "peerDependencies": {
-    "preact": "^7.1.0",
-    "preact-compat": "^3.13.1",
-    "prop-types": "^15.5.8",
-    "fabric-scss": "git+https://github.com/fabric-design/scss.git#master"
   }
 }
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "eslint-plugin-jasmine": "2.2.0",
     "eslint-plugin-jsx-a11y": "4.0.0",
     "eslint-plugin-react": "6.10.3",
+    "fabric-scss": "git+https://github.com/fabric-design/scss.git",
     "fs-then-native": "2.0.0",
     "glob": "7.1.1",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -101,11 +101,5 @@
     "webpack-node-externals": "1.5.4",
     "wholesale-styleguide-scss": "git+https://github.com/wholesale-design-system/scss.git#master",
     "yargs": "7.0.2"
-  },
-  "peerDependencies": {
```
fragsalat: Pls install all dependencies such React and Preact only as peer. Otherwise NPM will install this as production dependencies even If I don't use it. If you use `npm i -D preact` it will install them as dev deps and you have to manually put them to peer.

szafranek: Good point. I'll look into it on Monday (I'm on vacation now).

BTW, what do you think about moving some other deps, such as sinon, eslint into `devDependencies`? They shouldn't be production dependencies either. (I would do such cleanup in a separate PR, since that's out of scope for this one.)

package.json
```diff
@@ -101,11 +101,5 @@
     "webpack-node-externals": "1.5.4",
     "wholesale-styleguide-scss": "git+https://github.com/wholesale-design-system/scss.git#master",
```
szafranek: This URL has to be updated, now that the Fabric projects have moved.

package.json
fragsalat: So have a nice time :)
Actually all dependencies should be installed with -D flag for devDeps. Only react, preact and fabric-scss should be installed as peerDependencies since this packages are optionally needed in production.

package.json
```diff
@@ -84,8 +84,8 @@
     "mocha-webpack": "0.7.0",
     "node-sass": "4.5.2",
     "pre-commit": "1.2.2",
-    "preact": "7.2.1",
-    "preact-compat": "3.14.3",
+    "preact": "8.2.5",
+    "preact-compat": "3.17.0",
```
fragsalat: PS: The reason the demo didn't worked with preact was the ws-tile-chart component.
@inanda imported react from the package react which is wrong. 
We have to use the imports.js which handles the switch between preact and react.
I fixed this here https://github.com/fabric-design/components/pull/123/commits/4304d172f949046e84d5164d0f8dee10a3cce29f

szafranek:
I restored `peerDependencies`. `preact`, `preact-compat`, `prop-types`, `react-dom`, `fabric-scss` have to be listed in BOTH `devDependencies` and `peerDependencies`:
* They are required in `devDependencies` during development, e.g. to make the `imports.js` work.
* They are needed in `peerDependencies` to indicate via npm warning message that *either* react or preact need to be installed for the components to work.

PR #127, opened 2017-09-07
Dropdown keyboard control

fragsalat:
To improve the usability I made the dropdown controllable with keyboard.
Use arrow down / up to focus the next or previous item.
Use enter to select / store the focused item.
For multi select's you still have to hit the OK button which can be done with tab + enter.

Note: There are still some small issues in when marking an item as selected and pressing down key again the focus jumps down to the previous selection. This is not critical but I have no more time to fix this.

As well I added unit tests for the most important functions.

@tonysaad Here you can see how I workaround the React event handling :)

### DIFF
--- a/src/ws-dropdown/dropdown-menu.js
+++ b/src/ws-dropdown/dropdown-menu.js
@@ -60,6 +60,7 @@ export class DropdownMenu extends Component {
   constructor(props, context) {
     super(props, context);
     this.openSubMenu = null;
+    this.selectedIndex = -1;
     this.state = {
       filter: props.filter,
       items: props.items,
@@ -73,7 +74,12 @@ export class DropdownMenu extends Component {
    */
   componentDidMount() {
     if (this.input) {
+      this.input.addEventListener('keyup', this.onKeyUpUpdateFilter);
       this.input.addEventListener('change', event => event.stopPropagation());
+    }
+    if (this.button) {
+      this.button.addEventListener('click', this.onClickSubmit);
+      this.button.addEventListener('keydown', event => event.stopPropagation());
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/ws-dropdown/ws-dropdown.js
```diff
@@ -185,11 +231,15 @@ export class WSDropdown extends Component {
...
    */
   createState(props) {
-    const state = {
-      text: this.getTextFromValue(props.value, props.text),
-      value: this.enrichItems(props.value),
-      items: this.enrichItems(props.items)
-    };
+    const items = this.enrichItems(props.items);
+    let value = props.value;
+    // Map value to the item to improve usability
```
szafranek: I was expecting to see `map` after reading the comment. Maybe it could be dropped completely, since the code is more precise about what's going on here.

src/ws-dropdown/ws-dropdown.js
```diff
@@ -246,31 +296,47 @@ export class WSDropdown extends Component {
...
+    } else if (WSDropdown.openDropdown) {
+      WSDropdown.openDropdown.close();
     }
-    this.opened = true;
+
+    WSDropdown.openDropdown = this;
     this.dropdownContainer.style.height = 0;
     this.dropdownContainer.classList.add('mod-open');
     this.adjustSize(this.dropdownMenu.getHeight());
+    // Bind keyboard listener for keyboard navigation
```
szafranek: Perhaps this comment can be omitted, since it doesn't say much more than the code.

src/ws-dropdown/ws-dropdown.js
```diff
@@ -246,31 +296,47 @@ export class WSDropdown extends Component {
...
+    WSDropdown.openDropdown = null;
     this.animateElement(this.dropdownContainer, 'animate-close', container => {
-      this.opened = false;
       container.classList.remove('mod-open');
       // If this a multi select dropdown abort
       if (this.props.multiple) {
         this.dropdownMenu.clearSelections();
       }
     });
+    // Unbind keyboard listener
```
szafranek: Perhaps this comment could be skipped, as the code is self-explanatory.

tests/ws-dropdown/dropdown-menu.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(dm.getItemAtIndex(0).item).toBe(items[0]);
+    expect(dm.getItemAtIndex(5).item).toBe(items[0]);
+    expect(dm.getItemAtIndex(-1).item).toBe(items[2]);
+
+    dm.state.value = [items[1]];
+    expect(dm.getItemAtIndex(0).item).toBe(items[0]); // index 0 starts in value list
+    expect(dm.getItemAtIndex(1).item).toBe(items[1]); // index 1 starts after value list
+    expect(dm.getItemAtIndex(5).item).toBe(items[0]); // out of bound leads to index 0 which is in value list
+    expect(dm.getItemAtIndex(-1).item).toBe(items[2]);
```
szafranek: To a Fabric newbie such as myself this is not very clear. In JavaScript `-1` usually refers to an index that doesn't exist in the array. Why is it equal to item 2?

This is a place where a short comment would be helpful :).

tests/ws-dropdown/ws-dropdown.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    const dd = new WSDropdown({type: 'button'});
+    spyOn(dd, 'close');
+
+    dd.onGlobalKeyDown({key: 'ArrowUp'});
+    expect(dd.close).not.toHaveBeenCalled();
+    dd.onGlobalKeyDown({key: 'Escape'});
+    expect(dd.close).toHaveBeenCalled();
+  });
+
+  it('get\'s text from value for selects', () => {
```
szafranek: get's -> gets

src/ws-dropdown/dropdown-menu.js
```diff
@@ -158,26 +292,11 @@ export class DropdownMenu extends Component {
           item.selected = false;
         }
       });
+      this.setState({items: this.state.items});
```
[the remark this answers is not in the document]
fragsalat: Actually I don't really like the immutable state principle so a this.forceUpdate() would work as well^^

src/ws-dropdown/ws-dropdown.js
fragsalat: Haha It was my last change which introduced the test error. Was in hurry and now you see the result^^

src/ws-dropdown/dropdown-menu.js
```diff
@@ -105,11 +111,86 @@ export class DropdownMenu extends Component {
...
+    }
   }
 
   /**
+   * Bind keyboard listener and focus input if available when dropdown opens
+   * @returns {void}
+   */
+  onOpen = () => {
+    if (this.input) {
+      this.input.focus();
```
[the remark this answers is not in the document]
fragsalat: This would lead to this.dropdownInput.input.focus() or DropdownInput has to create a proxy function to call focus

src/ws-dropdown/ws-dropdown.js
fragsalat: I would change it to a more descriptive one => "For better usability the value can be a primitive value matching a dropdown item value"

tests/ws-dropdown/dropdown-menu.spec.js
fragsalat: Imagine you have selected the index 0 and press the arrow up key, what would you expect to happen? Probably that the selection jumps to the end of the list.
So index 0 - 1 is out of bounds and will go to last item. Therefore -1 is 2 :)

src/ws-dropdown/dropdown-menu.js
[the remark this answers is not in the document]
fragsalat: I would say fixing this is a bit more effort so I would skip this since it works and the immutable state is not really necessary.

tests/ws-dropdown/dropdown-menu.spec.js
szafranek: Makes good sense, thanks. I added a comment here.

szafranek:
👍

PR #134, opened 2017-10-13
fix(header): Removed refresh token logic

fragsalat:
I removed the refresh token logic since it is not anymore supported by IAM.

### DIFF
--- a/src/ws-header/authorization.js
+++ b/src/ws-header/authorization.js
@@ -6,18 +6,14 @@ export class Authorization {
   /**
    * @param {AbstractStorage} storage Key value storage
    * @param {string} loginUrl Url the user get's redirected to authorize
-   * @param {string} refreshUrl Url the app will send a POST to request a new access token
    * @param {string} clientId OAuth2 client id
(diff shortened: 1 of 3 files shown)


src/ws-header/authorization.js
```diff
@@ -89,14 +65,13 @@ export class Authorization {
...
-   * @param {Object} params Response parameters containing access and refresh token
+   * @param {Object} params Response parameters containing access token
    * @returns {void}
    * @private
    */
   updateTokens(params) {
     const expires = params.expires_in ? parseInt(params.expires_in, 10) : 3600;
     this.storage.set('access_token', params.access_token);
-    this.storage.set('refresh_token', params.refresh_token);
     this.storage.set('expires_at', new Date().getTime() + expires * 1000);
```
fokusferit: Is it still expiring that early ? ^^

fragsalat: It's expiring after one hour but thanks to google auth and this login implementation it's not hard to receive a new token.

PR #136, opened 2017-10-17
Tab menu

fragsalat:
I implemented the tab menu as stated in invision.

https://zalando.invisionapp.com/d/main#/console/10454474/256823763/preview

image


fokusferit:
First thing, you should really try to have atomic PR's as I see a fix(header) and feat(tab-menu). We should try to stick to our contributing guides. Even for small things.

tests/ws-tab-menu/ws-tab-menu.spec.js
```diff
@@ -0,0 +1,76 @@
...
+import {TestComponent} from '../test-component';
+
+function getExpectedBounds(tabItem) {
+  return {
+    left: tabItem.offsetLeft,
+    right: tabItem.parentNode.offsetWidth - tabItem.offsetLeft - tabItem.offsetWidth
+  }
+}
+
+describe('A WSTabMenu', () => {
```
fokusferit: +1 for having tests

src/ws-tab-menu/ws-tab-menu.js
```diff
@@ -0,0 +1,166 @@
...
+    );
+  }
+
+  /**
+   * Render the component markup
+   * @returns {XML}
+   */
+  render() {
+    return (
+      <div className="tab-menu-container" ref={element => { this.element = element; }}>
```
fokusferit: is this really needed ? I'm not seeing a usage in the ws-tab itself

fragsalat: Yes. Have a look at the Component.dispatchEvent.

fragsalat:
If you check the files there is no change in the headers. This is because squash commits creates new commit hashes which are shown here as new commits.

fragsalat:
And yes the tab menu only implements the tab menu. It propagates the changed value and the rest is left to you application

.gitignore
```diff
@@ -38,3 +38,5 @@ node_modules
 .idea
 **/*.iml
 styleguide.css
+
+package-lock.json
```
fokusferit: Isn't the purpose of a package-ock.json to put it into the repo to ensure same install on any machine ?

src/ws-tab-menu/ws-tab-menu.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+  }
+
+  /**
+   * Update state from property change
+   * @param {Object} props React properties
+   * @returns {void}
+   */
+  componentWillReceiveProps(props) {
+    this.setState(this.createState(props));
```
fokusferit: From my humbled understanding should not rely directly from props as both are updated async (https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) the save thing (we are also using in our new project) is to use an arrow function instead especially as on the next line we do a check with this.state.value and this might be another value then the one you save with setState ?

fragsalat: This problem only appears if I rely on an old state to calculate the new one. But in this component I just take the changed props and generate a completely new state. Therefor I would say it is ok. If I'm wrong or I misunderstood something I would be happy to talk about a better solution :)

fragsalat:
thx for the review. Lets talk about the package-lock.json tomorrow

PR #141, opened 2017-10-26
fix(header): Fixes useless token change events [BREAKING CHANGE]

fragsalat:
This PR fixes that the header fires ws-auth-change events even if the accessToken didn't changed.

*Important* This introduces a breaking change in the getAccessToken function which won't return a Promise anymore.

Before:
```js
WSHeader.getAccessToken().then(token => doSomething(token));
```

After:
```js
doSomething(WSHeader.getAccessToken());
```


src/ws-header/authorization.js
```diff
@@ -31,8 +26,11 @@ export class Authorization {
    * @returns {void}
    */
   changeAccessToken(accessToken) {
```
fokusferit: could you just change the docs for that ? looks like you now expect a string or a function as input ?

src/ws-header/authorization.js
```diff
@@ -65,7 +63,7 @@ export class Authorization {
 
   /**
    * Update the tokens and notify the header
-   * @param {Object} params Response parameters containing access token
+   * @param {{expires_in: number, access_token: string}} params Response parameters containing access token
```
fokusferit: +1

src/ws-header/ws-header.js
```diff
@@ -139,14 +153,9 @@ export class WSHeader extends Component {
...
-    this.authorization = new Authorization(
-      WSHeader.storage,
-      props.loginUrl,
-      props.clientId,
-      props.businessPartnerId
-    );
+    this.constructor.authorization = this.constructor.authorization || new Authorization(WSHeader.storage);
     // Listen to authorization changes
-    this.authorization.onAccessTokenChange(accessToken => {
+    this.constructor.authorization.onAccessTokenChange(accessToken => {
```
fokusferit: Why do you need constructor here ?

src/ws-header/ws-header.js
```diff
@@ -156,14 +165,14 @@ export class WSHeader extends Component {
       this.dispatchEvent('ws-auth-changed', accessToken);
     });
     // Check if we was redirected from the auth page and an access token is available
-    this.authorization.tryFetchToken(location.hash.substr(1));
+    this.constructor.authorization.tryFetchToken(location.hash.substr(1));
     // Listen for authentication requests
     window.addEventListener('ws-authorize', () => {
-      this.authorization.authorize();
+      this.constructor.authorization.authorize(props.loginUrl, props.clientId, props.businessPartnerId);
```
fokusferit: Where does this businessPartnerId comes from ? Are we allowed to share this here publicly ? I don't want to ran into any sort of security / compliance issue as we expose our IAM endpoint + some id

src/ws-header/ws-header.js
fragsalat: authorization is a static member

src/ws-header/ws-header.js
fragsalat: Actually I don't use PropTypes and just mock the functions. So PropTypes.string.isRequired would lead to an undefined error for me.

fokusferit: ok, are you handling such error ? E.g. in authorization.js -> buildQuery throwing some Error ? 😄  or `console.error` so people get immediate feedback ? I know it is not necessary as rendering and using it will throw an error anyway

PR #145, opened 2017-11-18
Remove unused Sinon.JS

mroderick:
This saves time and bandwidth during each install, which will make builds go faster.

#### Before

```text
added 1496 packages
```

#### After

```text
added 1492 packages
```

#### To verify

1. Check out this branch
1. `rm -rf node_modules`
1. `npm install`
1. `npm test`


mroderick:
Thanks

PR #146, opened 2017-11-22
ws-header show optional locale and oauth

antnascimento:
Component Screenshot:
image

Description (short): 
In case someone wants to use the header without displaying login and locale settings there are 2 properties to hide it now. For hiding locale on the header use showLocale={false} and to hide login use showAuthorization={false}. By default these are show.
Also a new property was created to set the rootUrl. By default the value is '#'.


jseb:
👍

PR #149, opened 2017-12-11
feat(header): Added a wrapper for the access token to check expiration of token

fragsalat:
Instead of return the access token as string and exposing many token related functions from the header I wrapped the token with a class.
This class allows to get the abbreviation of the user the token was issued for and to check if the token is valid based on the expiration date and JSON structure.

```js
WSHeader.storage.set('access_token', 'some.jwt.token');
const token = WSHeader.getAccessToken()
token.isValid() // Returns true if token is not expired
token.getUserAbbreviation() // Returns abbreviation of user the token was issued for e.g. tschlage
token == 'some.jwt.token' // will be true because valueOf returns the token
token === 'some.jwt.token' // will be false because the type is still JsonWebToken
`The token is: ${token}` // The JsonWebToken class implements a toString method
```

### DIFF
--- a/src/ws-header/access-token.js
+++ b/src/ws-header/access-token.js
@@ -1,0 +1,75 @@
+
+const log = console;
+/**
+ * This class represents a JWT token
+ */
+export class JsonWebToken {
+
+  /**
+   * Set the token
+   * @param {string} token Json Web Token
+   */
+  constructor(token) {
+    this.token = token;
+  }
+
+  /**
+   * Parse the token into an object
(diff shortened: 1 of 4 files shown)


src/ws-header/access-token.js
```diff
@@ -0,0 +1,75 @@
+
+const log = console;
+/**
+ * This class represents a JWT token
+ */
+export class JsonWebToken {
```
fokusferit: 👍

PR #150, opened 2017-12-12
fix(header): Fixed file name and added JSON serializer function to token class

fragsalat:
- Fixed naming of file access-token.js to json-web-token.js
- Fixed serialization of token by adding a toJSON method to the JsonWebToken


fokusferit:
👍

PR #158, opened 2018-04-25
React components

mousemke:
this PR relates to work in LBJ-716 and usage of fabric as a react component library

in addition to this specific ticket, this bundle makes fabric able to be embedded into any react (and preact?) project while trying to minimize tech requirements. This will also make it easy to publish fabric to npm

this is still a work in progress since refactoring of the scss files will dramatically reduce the bundle size


mousemke:
👎

fragsalat:
First of all we already use gulp so I don't got the point why you added webpack now?^^
If I got it right you are bundling all the css of fabric scss and components into the index.js so regardless if I need the scss or not I must load it?
When this is true why you don't add those webpack rules and importer to you project but to this?

mousemke:
That's fair, but the project does use webpack, though merely for testing. This utilizes a webpack config that's already there, just not being used. The intention of this PR (When finished) would be that all previously available use cases and functionality is unchanged. This only adds an additional file with the library bundled together.  The bonus of this is that it is zero config. No scss. No setup. Just require.

fragsalat:
Ah ok I was confused because of the added webpack loaders. In general I would be fine with it but our build process for releasing new versions looks like this
```bash
# On every push
gulp build
gulp lint
gulp test
# On push to master only (PR merged)
gulp generate-docs bump-version changelog
```

src/ws-header/ws-header.js
```diff
@@ -287,9 +287,16 @@ export class WSHeader extends Component {
                   ref={element => { this.menuItems[index] = element; }}
                   className={(link.isCurrent) ? 'is-current' : null}
                 >
-                  <a href={link.href} onClick={event => { if (link.onClick) link.onClick(event); }}>
-                    {link.label}
-                  </a>
+                {
+                  (
+                    link.$$typeof ? link :
```
fragsalat: In general it's ok but pls keep the formatting as it was.
```
{link.$$typeof ?
  link
:
  <a href={link.href}>
    {link.label}
  </a>
}

Comment (fragsalat) [reply src/ws-header/ws-header.js]: If you would add this in a standalone pr we could merge it

Comment (mousemke) [reply src/ws-header/ws-header.js]: 👍

**Comment by mousemke** [comment]
@fragsalat new PR opened with only that fix.  side note on formatting - i noticed there was a prettier ticket open.  i highly recommend it.  it removes the whole discussion ^^

**Comment by mousemke** [comment]
@fragsalat two questions:
+ how would you feel about throwing all the gulp commands into one npm command? 
    + it's ~~easier~~ safer than running command line things from gulp
+ if the dist files are generated during release, can we add them to `.gitignore`?

**Comment by fragsalat** [comment]
Lets talk about this when i'm back at work. I'm not able to think about this right now.

**Comment by mousemke** [comment]
👍

File: dist/amd/ws-header/ws-header.js [hunk 2]
```diff
@@ -148,7 +148,7 @@ define(['exports', '../imports', './storage/cookie-storage', './storage/local-st
           _this2.dispatchEvent('ws-auth-changed', accessToken);
         });
 
-        this.constructor.authorization.tryFetchToken(location.hash.substr(1));
+        this.constructor.authorization.tryFetchToken(window.location.hash.substr(1));
```
fragsalat: Don't include dist folder in your PR pls. The dist folder and doc folder is auto generated by a circleci process

src/imports.js
```diff
@@ -22,7 +23,7 @@ export class Component extends _React_.Component {
    * @returns {void}
    */
   dispatchEvent(name, detail, bubbles = true) {
-    const event = new CustomEvent(name, {detail, bubbles});
+    const event = new window.CustomEvent(name, {detail, bubbles});
```
fragsalat: Why did you add the window here?

src/imports.js
```diff
@@ -2,6 +2,7 @@ import * as _React_ from 'react';
 import * as ReactDOM from 'react-dom';
 import * as types from 'prop-types';
 
+
```
fragsalat: Why a additional line is needed?

src/ws-header/storage/local-storage.js
```diff
@@ -13,7 +13,7 @@ export class LocalStorage extends AbstractStorage {
    */
   set(key, value) {
     const encodedValue = encodeURIComponent(JSON.stringify(value));
-    localStorage.setItem(`${this.name}${key}`, encodedValue);
+    window.localStorage.setItem(`${this.name}${key}`, encodedValue);
```
fragsalat: Why do you add this obsolete window object here?

src/ws-header/ws-header.js
```diff
@@ -259,6 +259,14 @@ export class WSHeader extends Component {
     this.level2.classList.remove('is-active');
   }
 
+  renderLink(link) {
```
fragsalat: Still JSDoc is missing

src/ws-header/ws-header.js
```diff
@@ -333,9 +343,9 @@ export class WSHeader extends Component {
             <ul className="main-sub-menu" key={`sub-menu${index}`} ref={element => { this.subMenus[index] = element; }}>
               {parent.children.map((child, childIndex) =>
                 <li key={`sub-link-${index}-${childIndex}`} className={(child.isCurrent) ? 'is-current' : null}>
-                  <a href={child.href} onClick={event => { if (child.onClick) child.onClick(event); }}>
-                    {child.label}
-                  </a>
+                {
```
fragsalat: Pls configure your formatter to keep the formatting like thise `{this.renderLink(link)}`

src/ws-header/ws-header.scss
```diff
@@ -139,5 +139,8 @@ $header-border-color: schema-color(4) !default;
   }
   a {
     color: inherit;
+    &:focus {
+      color: #ffffff;
```
fragsalat: Pls use the variable $white instead

webpack.config.preact.test.js
```diff
@@ -1,35 +1,8 @@
 const path = require('path');
+const testBase = require('./webpack.config.test-base');
 
 module.exports = {
-  devtool: 'inline-source-map', // just do inline source maps instead of the default
```
fragsalat: I likes the old formatting way more^^

dist/amd/ws-header/ws-header.js
mousemke: hehe agreed https://github.com/fabric-design/components/pull/158#issuecomment-384986699

webpack.config.preact.test.js
mousemke: me too :(

this was done to try to make codeclimate happy

src/ws-header/ws-header.js
mousemke: 👍

src/imports.js
mousemke: backwards compatibility.  there isn't always a CustomEvents global (and / or environments dont always default to checking the `window` object for globals; node test environments break)

screen shot 2018-05-16 at 14 38 16

src/ws-header/storage/local-storage.js
mousemke: see above

src/imports.js
fragsalat: I never had this error before so there must be a change in meanwhile. Even if there is the window object now the error must be kind of smae because CustomEvent will be undefined then

src/ws-header/ws-header.js
```diff
@@ -287,9 +300,7 @@ export class WSHeader extends Component {
                   ref={element => { this.menuItems[index] = element; }}
                   className={(link.isCurrent) ? 'is-current' : null}
                 >
-                  <a href={link.href} onClick={event => { if (link.onClick) link.onClick(event); }}>
-                    {link.label}
-                  </a>
+                {link.$$typeof ? link: this.renderLink(link)}
```
fragsalat: Missing indention

src/ws-header/ws-header.js
```diff
@@ -333,9 +344,7 @@ export class WSHeader extends Component {
             <ul className="main-sub-menu" key={`sub-menu${index}`} ref={element => { this.subMenus[index] = element; }}>
               {parent.children.map((child, childIndex) =>
                 <li key={`sub-link-${index}-${childIndex}`} className={(child.isCurrent) ? 'is-current' : null}>
-                  <a href={child.href} onClick={event => { if (child.onClick) child.onClick(event); }}>
-                    {child.label}
-                  </a>
+                {this.renderLink(child)}
```
fragsalat: Missing indention

mousemke:
@fragsalat 

> I never had this error before so there must be a change in meanwhile. Even if there is the window object now the error must be kind of smae because CustomEvent will be undefined then

i dont understand what you mean here.

PR #160, opened 2018-04-27
feat(header) Accept custom link elements

mousemke:
for #167 

Pull request for issue: LBJ-716

links arent always simple anchors.  in my example i needed a `<Link />` element (react router link) instead of an `<a />`.  this will detect something that isn't a standard object and render it as a component instead

### DIFF
--- a/src/ws-header/ws-header.js
+++ b/src/ws-header/ws-header.js
@@ -273,6 +273,19 @@ export class WSHeader extends Component {
   }
 
   /**
+   * codeclimate demands abstraction; this makes links
+   * @param  {Object} link link data, label, href, etc
+   * @return {JSX} a rendered link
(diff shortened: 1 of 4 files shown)


src/ws-header/ws-header.js
```diff
@@ -259,24 +259,34 @@ export class WSHeader extends Component {
     this.level2.classList.remove('is-active');
   }
 
+  renderLink(link) {
```
fragsalat: JSDoc missing :)

mousemke: when #158 is merged this will be fixed

mousemke:
👎

mousemke:
waiting until the rollback is committed

PR #165, opened 2018-05-18
chore(repo) - removed dist and doc

mousemke:
for #166 

removed dist and doc directories and added them to `.gitignore`

this comes from discussions about not committing updated dist files since they're generated. and not removing them fore the sake of a very dirty diff.  if we do this in one place, we can keep the dirty diff all in one place


fragsalat:
As I said. If they are in gitignore they are not included in the release. Just don't add them in your commit please

mousemke:
ah.  misinterpreted that.  surely there's a way to fix that aside from `checkout` than all the time, but not super important

Issue #163, opened 2018-05-15
Use setState in WSDropDown


anstit:
> Warning: DropdownMenuItem.componentWillReceiveProps(): Assigning directly to this.state is deprecated (except inside a component's constructor). Use setState instead.

I am getting this Warning in my Browser each time the page is loaded for the first time.

mousemke:
(closed but i cant close the issue myself)

PR #170, opened 2018-05-18
fix(dropdown) no longer directly accesses the state

mousemke:
for #163 

we no longer access the state directly; now using setState

### DIFF
--- a/src/ws-dropdown/dropdown-menu-item.js
+++ b/src/ws-dropdown/dropdown-menu-item.js
@@ -77,7 +77,7 @@ export class DropdownMenuItem extends Component {
    * @returns {void}
    */
   componentWillReceiveProps(props) {
-    this.state = props.item;
+    this.setState(props.item);
   }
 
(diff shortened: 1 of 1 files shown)


src/ws-dropdown/dropdown-menu-item.js
```diff
@@ -113,16 +113,20 @@ export class DropdownMenuItem extends Component {
         if (this.state.selected) {
           this.props.handle('change', null);
         } else {
-          this.state.selected = true;
-          this.state.stored = true;
+          this.setState({
+            ...this.state,
```
fragsalat: I thought setState is merging the state anyways so this line would be obsolete or am I wrong?

mousemke: yes.  i initially did it just for compat vs. the original, but i think you're right.  i'll do a quick check then update

mousemke:
:+1:

PR #173, opened 2018-05-23
misc(build) lint fixes

mousemke:
many lint fixes to fix the build post-dependency update

### DIFF
--- a/src/ws-header/ws-header.js
+++ b/src/ws-header/ws-header.js
@@ -332,27 +332,31 @@ export class WSHeader extends Component {
                 </li>
               }
               {this.props.showAuthorization && (!this.state.isLoggedIn ?
-                <a // eslint-disable-line jsx-a11y/anchor-is-valid
-                  role="button"
-                  onClick={() => this.login()}
-                  onKeyPress={e => {
-                    if (e.key === 'enter') this.login();
-                  }}
-                  tabIndex={0}
-                >
-                  <li>Login</li>
-                </a>
+                <li>
+                  <a // eslint-disable-line jsx-a11y/anchor-is-valid
+                    role="button"
+                    onClick={() => this.login()}
+                    onKeyPress={e => {
+                      if (e.key === 'enter') this.login();
+                    }}
+                    tabIndex={0}
+                  >
+                    Login
+                  </a>
+                </li>
                 :
-                <a // eslint-disable-line jsx-a11y/anchor-is-valid
-                  role="button"
-                  onClick={() => this.logout()}
-                  onKeyPress={e => {
-                    if (e.key === 'enter') this.logout();
-                  }}
-                  tabIndex={0}
-                >
-                  <li><span className="icon icon24 icon-power" /></li>
-                </a>
+                <li>
+                  <a // eslint-disable-line jsx-a11y/anchor-is-valid
+                    role="button"
+                    onClick={() => this.logout()}
+                    onKeyPress={e => {
+                      if (e.key === 'enter') this.logout();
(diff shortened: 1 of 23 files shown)


src/ws-inline-edit/ws-inline-edit.js
```diff
@@ -100,7 +91,13 @@ export class WSInlineEdit extends Component {
    */
   render() {
     return (
-      <div className="ws-inline-edit" onClick={() => this.editElement()}>
+      <div
+        className="ws-inline-edit"
+        onClick={() => this.editElement()}
+        onKeyPress={e => {
```
fragsalat: I would say a keyPress on a non selectable element makes no sense or?
I never can tab to this element until a tabindex is making it focusable. Does a tabindex make it focusable even?^^

src/ws-header/ws-header.js
```diff
@@ -336,13 +332,27 @@ export class WSHeader extends Component {
...
+                <a // eslint-disable-line jsx-a11y/anchor-is-valid
+                  role="button"
+                  onClick={() => this.login()}
+                  onKeyPress={e => {
+                    if (e.key === 'enter') this.login();
+                  }}
+                  tabIndex={0}
+                >
+                  <li>Login</li>
+                </a>
```
fragsalat: This is not valid HTML

src/ws-week-picker/ws-week-picker-calendar.js
```diff
@@ -108,18 +108,23 @@ export class WSWeekPickerCalendar extends Component {
           }
           const {week, year} = weekInMonth;
           return (
-            <td
+            <a // eslint-disable-line jsx-a11y/anchor-is-valid
```
fragsalat: This is not valid HTML. tr only can contain a td

mousemke:
if this isnt going to merge anyways, i wont bother with these fixes, however in an attempt to not go backwards in supporting a11y (which was mostly the fixes in the linter), i have moved warehouse-alloctation-ui to use `git+https://github.com/mousemke/components.git#masterWithDist`

1) tabindex makes things focusable so they can accept keypress events.
2) it does make it more useful. this is not the end all of support for a11y, and it's not all we can do, but it is a step in the right direction
3) though i was literally just trying to get the build to pass here, adhering our rules, it's worth noting that the rules didnt change.  clearly this is more of a case of the older version (for whatever reason) didnt actually enforce the rules that we gave it
4) invalid html rules i will fix, but again - on my own branch

src/ws-inline-edit/ws-inline-edit.js
fragsalat: But there is no tabindex or I'm just not seeing it? Anyways I think adding the tabindex 0 isn't right but I have to re-check it.

src/ws-option-buttons/ws-option-buttons.js
```diff
@@ -158,18 +157,18 @@ export class WSOptionButtons extends Component {
   render() {
     return (
       <div className="ws-option-buttons" ref={element => { this.element = element; }}>
-        {this.state.items.map((item, index) =>
+        {this.state.items.map((item, index) => (
           <div className={`option-button ${index < this.state.visible ? '' : 'is-hidden'}`}>
-            <a
+            <a // eslint-disable-line jsx-a11y/anchor-is-valid
```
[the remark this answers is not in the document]
fragsalat: I would not just use rules because others say that they are right and useful. When they don't apply to our code we don't need to disable them on every error manually but can disable it completely

PR #177, opened 2018-06-18
Multiselect

fragsalat:
- I implemented a new component called multi-select which extends the dropdown but with pulling out the filter input and selected values for more usability when having enough space. It was required by UX in Purchase Order frontend
- I cutted out the overlay logic from the dropdown to be able to reuse it for the coming new components like tooltip, new date picker etc.
- I fixed the linter errors after updating the dependencies
- I made some fixes to the dropdown and option buttons and the datepicker (when used in table header)

PS: Sry for the big PR but It was a production branch a long time

Examples of new component:
Selection: 
screenshot from 2018-06-18 18-01-03
Values: 
screenshot from 2018-06-18 18-01-13


src/ws-dropdown/ws-dropdown.scss
```diff
@@ -0,0 +1,255 @@
+$export-dropdown-css: true !default;
```
fokusferit: I think it makes sense to release this to fabric-design/scss

fragsalat: Actually I moved it from scss to this place as the dropdown is only used together with the component and I guess no one will ever implement it's own logic of it.

src/ws-date-picker/ws-date-picker.js
```diff
@@ -67,7 +68,8 @@ export class WSDatePicker extends Component {
       onChange: this.onChange.bind(this)
     });
     // Prevent default change event bubbling
-    this.input.addEventListener('change', event => event.stopPropagation(), true);
+    this.input.addEventListener('change', this.stopPropagation);
```
fokusferit: was there a bug or need for updating this ?

src/ws-dropdown/dropdown-input.js
```diff
@@ -35,7 +35,7 @@ export class DropdownInput extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      value: props.value
+      value: props.value && props.value.value
```
fokusferit: is it intended to have a possible 'undefined' as a state value ? Seems a bit unsafe

src/ws-dropdown/dropdown-menu-item.js
```diff
@@ -96,40 +108,44 @@ export class DropdownMenuItem extends Component {
...
+    } else if (!this.context.multiple) {
+      // If it is selected we publish null because it will be deselected in the upper menu
+      if (item.selected) {
+        this.props.handle('change', null);
       } else {
-        this.setState({selected : !this.state.selected});
+        item.selected = true;
+        item.stored = true;
+
+        this.setState({item});
```
fokusferit: shouldn't this be with a callback and using prevState? Basically what you do here is
`this.setState({item: this.state.item})` as you reference it at the top, right ?

src/ws-dropdown/dropdown-menu-item.js
```diff
@@ -96,40 +108,44 @@ export class DropdownMenuItem extends Component {
...
+
+        this.props.handle('change', item);
       }
+    } else {
+      item.selected = !item.selected;
+      this.setState({item});
```
fokusferit: same here I would say.

src/ws-dropdown/dropdown-menu.js
```diff
@@ -345,7 +338,10 @@ export class DropdownMenu extends Component {
         this.showChild(data);
         break;
       case 'change':
-        this.clearSelections();
+        // Clear filter input after selecting item
+        if (this.props.filterable) {
```
fokusferit: not necessary, but you changed everything to this.state.filterable, I guess also here?.

src/ws-dropdown/ws-dropdown.scss
fokusferit: Funny fact: We had to do it in Retail Portal as it was easier for the team to implement dropdown on it's own then integrating wsDropdown (which is mainly because they had concerns after not knowing how to easily start...). 
I'm still for separation, as there are even 1-2 teams outside retail core who are only using SCSS library, even if we never thought about this before, people start using scss more than the components (pattern lib) we have.

fragsalat: I'm fine with moving all the css to fabric-scss when we all align on it :)

src/ws-date-picker/ws-date-picker.js
fragsalat: The reason was simply a memory leak. Elements which still are referenced like with a event listener will not be cleaned up by the garbage collector. If you do addEventListener with a arrow or anonymus function you can not remove the listener by just applying this arrow function again as this is a new instance with a new identifier. Therefore I moved this into the class so that I can reference the exactly same function for adding and removing the listener

src/ws-dropdown/dropdown-input.js
fragsalat: Yes it's intended to have possibly an undefined as you can include a dropdown without having a default value or a previously persisted value. 
Would a ternary be better for you? Like
```js
this.state = {
  value: props.value ? props.value.value : null
}
```

src/ws-dropdown/dropdown-menu-item.js
fragsalat: Yes, the item is still referenced and I just call the setState fn to tell Preact/React to re-render.
What is the benefit from calling setState with a callback?

src/ws-dropdown/dropdown-menu-item.js
fragsalat: It's true that this.state probably is already modified but I don't see a problem here as the component is rendered only when calling setState. Or did I miss something?

src/ws-dropdown/dropdown-menu-item.js
fragsalat: I read about this asynchonous problem with setState but here it is in my eyes definitely not a problem. They say that you should not rely on this.state after calling setState and I definitely don't use this.state from the beginning of this function until the beginning of the render function, which indeed is called by setState. So no need to introduce more complex code here.

src/ws-dropdown/dropdown-menu.js
fragsalat: There is a tiny difference between filterable and filtered which is mentioned in the props description but maybe not clear here.
filterable -> Means that a dropdown has a filter input
filtered -> Means that the dropdown items are filtered but there is no filter input in the dropdown

The reason for this is that the ws-multi-select extends the dropdown but lays out the values and filter input. So filterable is true when <ws-dropdown filterable/> and should mean that the dropdown shows and cares about the filter value. Filtered is true when <ws-multi-select> and should mean that the filter logic of the dropdown is used but the multi select shows and cares about the filter value.

To make this more clear I'll add a more sophisticated comment here.

Issue #179, opened 2018-06-20
bug(date-picker) icon removes clickability


mousemke:
you cant click through the icon, so if you happen to hover over it you lose both the pointer and the ability to open the cal. view.

solution:
pointer events none on the icon

PR #180, opened 2018-06-20
fix(datepicker) disabled pointer events on the icon

mousemke:
now you can click through to the input

Pull request for issue: #179 

Component Screenshot:

screen shot 2018-06-20 at 17 56 07

### DIFF
--- a/src/ws-date-picker/ws-date-picker.scss
+++ b/src/ws-date-picker/ws-date-picker.scss
@@ -2,6 +2,7 @@
   position: relative;
   display: inline-block;
   &.with-input .icon {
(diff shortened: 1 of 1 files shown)


src/ws-date-picker/ws-date-picker.scss
```diff
@@ -2,6 +2,7 @@
   position: relative;
   display: inline-block;
   &.with-input .icon {
+    pointer-events: none;
```
fragsalat: I din't know this yet. I absolutely like this!!!! Before I always created additional listeners for the icons^^

PR #182, opened 2018-06-25
Fix datepicker pass class name

mousemke:
the date picker now accepts the className attribute for passing in extra classes - in my use case, an error class

normal:
screen shot 2018-06-25 at 16 27 12

error:
screen shot 2018-06-25 at 16 27 33

markup:
screen shot 2018-06-25 at 16 30 55


mousemke:
nope sorry.   forgot about the other commits that arent in this branch

src/imports.js
```diff
@@ -6,14 +6,14 @@ import * as types from 'prop-types';
...
-export const PropTypes = types;
-export const render = ReactDOM.render;
-
 /**
- * This is a abstract React/Preact component class
+ * Dispatches a native event
+ * @param {string} name Event name
+ * @param {*} detail Event detail data
+ * @param {boolean} bubbles True if the event bubbles up the dom tree
+ * @returns {void}
```
fragsalat: Why did you copy the comment from the dispatchEvent function to the class?

mousemke: hmm....   dont know where that came from.  guess it really doesnt matter though does it

PR #184, opened 2018-06-26
feat(datePicker) the date picker now accepts an onClick function

mousemke:
the date picker now accepts an onClick function.  seems generally useful but for example i'm using it to clear the error class when something is selected


fragsalat:
Whats the use case for this?

fragsalat:
Could you update the docs in components/doc/components/ws-datepicker.md and add the use case for onClick with example and description?

PR #187, opened 2018-07-04
fix(lint): fixing master branch

fokusferit:
Description (short): 

I can't move forward with having a huge PR which is also failing and fixes a lot of things + introduces a new component.

-> This will definitely create more bugs and headache for all of us. 

So I take the opportunity again to only fix ESLint Errors, so that CI is at least back to normal. No Doclint warnings, as they are now warnings.

Things I did here:

- I tried to use onKeyPress Only and for almost all cases it shouldn't make a big difference for now, as many onClick Handlers are not relying on the Event Object.
- I had to disable two lines because in one case rewrite is too much effort and second is just ESLint complaining about undef-state which is not true (but it can't understand the state init function).
- I only fixed Errors not Warnings, so the PR is smaller and easier to review. 

I did not change any rule for now, even when I thought some might make no sense for me 
-> This should be a group discussion.
-> I also have to admit I was a bit surprised that we had duplicate code, non defined PropTypes for defaultProps etc. => I think this is a result of not fixing CI for master and just merging over the last weeks.

I hope this gets us back to normal mode

### DIFF
--- a/src/ws-notification/ws-notification.js
+++ b/src/ws-notification/ws-notification.js
@@ -18,7 +18,6 @@ const DEFAULT_NOTIFICATION_TYPE = 'info';
  * }});
  */
 export class WSNotification extends Component {
-
   /**
    * @constructor
    */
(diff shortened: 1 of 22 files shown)


src/ws-notification/ws-notification.js
```diff
@@ -94,7 +95,7 @@ export class WSNotification extends Component {
    * @returns {void}
    */
   animateIn(notification, index) {
-    const list = this.list;
+    const {list} = this.list;
```
fragsalat: Just recognized while fixing the tons of merge conflicts. const list = this.list can not be replaced by a destruction in this way

PR #189, opened 2018-07-17
Added a working inline edit

fragsalat:
I added a inline edit based on the mockups used for the PO frontend.
Right now it supports the types text, number ad price. Those types also have a validation included.

More information you can find in the documentation file.

Initial state:
screenshot from 2018-07-17 15-24-38
Edit
screenshot from 2018-07-17 15-24-56
Edit in table
screenshot from 2018-07-17 15-25-14


src/ws-inline-edit/types/number-strategy.js
```diff
@@ -0,0 +1,81 @@
...
+      NumberStrategy.getPattern('de').test(value);
+  }
+
+  /**
+   * Converts the localized string number to a decimal
+   * @param {string} value The input value
+   * @returns {number}
+   */
+  convert(value) {
+    let simple;
```
jordi-rubio: This method fails if value is not set

jordi-rubio:
👍

PR #192, opened 2018-07-27
Fixed not changeable filter inputs

fragsalat:
The filter inputs of dropdown and multi select were not changeable anymore due to diffences of defaultValue and value in React.


tonysaad:
👍

PR #193, opened 2018-08-01
fix(overlay): Fixed broken components when blurring the tab

fragsalat:
**Explanation of the error:**
A component opens the dropdown when a input got focus and closes the dropdown when blurred. The close works by adding a css class to animate the element and a listener removes mod-open class when the animation is done. When you focus the input and the dd is opened and then switch the tab, chrome and ff will trigger a blur event, the animation is started but actually chrome and ff are not rendering anything. When you then focus the original tab again, the input will be focused and the dropdown is opened and finally chrome renders the close animation^^ At this point the dropdown thinks it's open but it isn't and it's broken. 

**Solution**
before starting the animation add another listener for window blurring to immediately call the callback like the animation would be done.


jordi-rubio:
👍

PR #194, opened 2018-08-01
fix(date-picker): Fixed picker not closing when clicking in another picker

fragsalat:
**Description:**
When opening one picker by clicking on the input and then opening another picker by clicking the input, the old picker doesn't close anymore. This happens because we catch click event to don't disturb other parent elements.

**Solution:**
To work around this I notify all flatpickr instances manually about the click and therefore close them.


jordi-rubio:
Don't you want to add a test?

tests/ws-date-picker/ws-date-picker.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable */
```
jordi-rubio: Do you want to rename the file to ws-date-picker-spec.js? ;)

PR #195, opened 2018-08-01
feat(dropdown): Added minFilterLength property to only show items when filter is long enough

fragsalat:
This allows the user to only show items when the filter value has a specific length. 

Requires min 2 chars in filter:
screenshot from 2018-08-01 18-05-16

Inserted 2 chars but nothing found
screenshot from 2018-08-01 18-05-34

Found items
screenshot from 2018-08-01 18-05-43


jordi-rubio:
When not filtering, wouldn't it make more sense to show the complete list of possible elements, so that the user can click them "manually"?
Otherwise it may be hard to filter something when the users don't know which values are possible...

fragsalat:
I argued the same but actually the use case of Basar Temur were different. I'll let him answer here

Chojaa:
The list is in our case too long thats why we need to "hide" the list and yes we can use limit but for example if we use limit 5 and we filter for  "Ne" we can see only 5 opportunities with "Ne" and not maybe all 10 which we need. Also possibly we are worry that the full list can confuse someone.

PR #197, opened 2018-08-22
fix(webpack) removes unused components from the bundle

mousemke:
this removes some packages from the webpack bundle that we are not using as well as switching the source of scss to the altered repo


mousemke:
not intended as a pr to `fabric-design` 

closing this

PR #198, opened 2018-08-24
fix(inline-edit): Resize inline edit when value changes from outside

fragsalat:
**Problem**
When the value property changed from outside the value of the input is updated but the width isn't recalculated.

**Solution**
Resize the input after the state change.


fokusferit:
👍

PR #202, opened 2024-01-29
Escape user input before creating RegExp

azimsky:
Un-sanitized user input causes SyntaxError when trying to create a regular expression. Example: try typing "("

Component Screenshot:
Screenshot 2024-01-26 at 13 26 08

Description (short): escaped special characters before using in RegExp


fragsalat:
:+1:

azimsky:
👍

